PR #302, opened 2015-03-23
Listen on fix

literatesnow:
ruby 1.9.3p194 (2012-04-20 revision 35410) [x86_64-linux]

Listen.on clients appear to be broken in 2.9.0 as hash keys for moved_from and moved_to don't exist in the smooshed hash (queue_optimizer.rb).

Server:
```
$ listen -d '.' -r -v -f 127.0.0.1:4500
```

Client:
```
require 'listen'

listener = Listen.on '127.0.0.1:4500' do |modified, added, removed, moved_from, moved_to|
  p modified
  p added
  p removed
  p moved_from
  p moved_to
end

listener.start
sleep
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ listen -v -d "/projects/my_project" -f "10.0.0.2:4000" # changes in given direct
 To connect to a broadcasting listener as a recipient, specify its address using `Listen.on`:
 
 ```ruby
-listener = Listen.on '10.0.0.2:4000' do |modified, added, removed|
+listener = Listen.on '10.0.0.2:4000' do |modified, added, removed, moved_from, moved_to|
   # This block will be called
 end
 listener.start
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ listen -v -d "/projects/my_project" -f "10.0.0.2:4000" # changes in given direct
 To connect to a broadcasting listener as a recipient, specify its address using `Listen.on`:
 
 ```ruby
-listener = Listen.on '10.0.0.2:4000' do |modified, added, removed, moved_from, moved_to|
+listener = Listen.on '10.0.0.2:4000' do |modified, added, removed|
   # This block will be called
 end
 listener.start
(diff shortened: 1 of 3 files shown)


e2:
@literatesnow - actually, the `move_to` and `move_from` were never meant to be part of the public API - they were for improving editor support on Linux only (they help make the events more accurate, e.g. when a file is saved when using backup files, a `:modified` event is correctly generated).

literatesnow:
@e2 - I see. In that case there could be a check to see if the change index appears in the smooshed hash to stop it from silently dying in ruby 1.9.3.

```
   end
  _squash_changes(_reinterpret_related_changes(cookies))
else
  smooshed = { modified: [], added: [], removed: [] }
  changes.each do |_, change, dir, rel_path, _|
    smooshed[change] << (dir + rel_path).to_s if smooshed.key?(change)
  end
  smooshed.tap { |s| s.each { |_, v| v.uniq! } }
```

e2:
@literatesnow - I restarted the build (the master branch of Celluloid was failing - sorry about that).

Yes - the above code should be fine (just add to the above and squash and it should be fine). 

Don't worry about the tests, since this is more of an architecture problem:

Lets say you have a slow drive, and an editor with lots of plugins, etc. When you "save" a file, there may be a few operations (writing to a temporary file, deleting the original file, then renaming the temporary file) - these operations generate more than one event on Linux (move_from, move_to, unlink, create, write, close, etc.).

The problem is: many people want the TCP adapter to be fast, so the packets are sent as soon as something happens. But because it happens immediately (as people want it), the server can't "optimize the queue" to realize that all those operations actually just "modfiy" the saved file (because it's deleted then created).

On the other hand, if the client optimizes the queue (as we have here), the files may not exist in the paths given (so the client cannot optimize the queue based on whether files actually exist or not).

That's why the `local_fs?` is there - just to avoid calling `File.exist?` when TCP is used.

So your original code would be "the best", because the client could use the `move_from` and `moved_to` to optimize the queue as well. But I can't assume that the list of files changed sent over TCP actually exist on the client. 

Listen does have a new `:relative` option, so maybe in the future Listen could more accurately report changes over TCP if the user enables checking for remote files locally.

literatesnow:
@e2 Thanks for the detailed explanation. I've made the changes as specified.

e2:
Thanks. I probably should release this.

PR #303, opened 2015-03-24
listen -d can take multiple directories

ideasasylum:
The `listen` script can now take multiple directories

I encountered the duplicate directories problem in my Rails project (a
vendor gem used symlinks) so the simplest solution was not to watch the
whole project directory (i.e., the default —directories option) but
just the directories I care about. That meant I needed to watch
multiple directories which listen supports but the `listen` script did
not.

This commit changes the `-d` option type from string to array and adds
the appropriate specs. The order of the options doesn’t matter: `listen
-d app spec -v -r` is the same as `listen -v -r -d app spec` and then
directories `app` and `spec` are pass to `Listen.to` as an array

### DIFF
--- a/lib/listen/cli.rb
+++ b/lib/listen/cli.rb
@@ -21,7 +21,7 @@ module Listen
                  banner:  'The address to forward filesystem events'
 
     class_option :directory,
-                 type:    :string,
+                 type:    :array,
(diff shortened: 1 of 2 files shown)


e2:
Thank you very much - and for the tests!

PR #305, opened 2015-04-11
Anticipate Celluloid::FSM breaking off into its own gem.

digitalextremist:
This will change to a static ( non-github ) rubygems reference, and is not intended to be merged right away. This is anticipating celluloid/celluloid#547, when `Celluloid::FSM` is no longer in `Celluloid` core, and is moved out into its own gem [celluloid-fsm](/celluloid/celluloid-fsm) ( currently in prerelease also ).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,7 @@ end
 
 group :test do
   gem 'celluloid', github: 'celluloid/celluloid', branch: '0-16-stable'
+  gem 'celluloid-fsm', github: 'celluloid/celluloid-fsm', branch: 'master'
   gem 'celluloid-io', '>= 0.15.0'
   gem 'rake'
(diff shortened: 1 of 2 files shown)


e2:
Thanks. `celluloid-pool` and `celluloid-supervisors` probably will also have to be added. But I think it would be best for Celluloid to temporarily contain these dependencies itself (to avoid breakage like this until 1.x).

digitalextremist:
I am beginning to think the same.

e2:
I duplicated the FSM code and adapted it to avoid referencing any Celluloid specifics or causing any conflicts. I'll watch celluloid-fsm in case of updates, bugs.

digitalextremist:
@e2 so far `celluloid-fsm` has probably been the least changed gem of the lot, it's pretty much LTS.

PR #318, opened 2015-06-12
Listen 3x proposal

e2:
See #319 for details

Commit messages are self-explanatory, including:

- removing TCP functionality
- removing Celluloid


lib/listen.rb
```diff
@@ -1,16 +1,27 @@
-require 'celluloid'
+require 'logger'
+require 'listen/logger'
 require 'listen/listener'
 
 require 'listen/internals/thread_pool'
 
+# Always set up logging by default first time file is requried
```
rymai: Typo spotted `requried` => `required`.

mechanicalduck:
Remove TCP functionality - does this mean that Listen cannot be used over network anymore?

rymai:
No, it's moved into two separate gems (see #319). ;)

lib/listen/listener.rb
```diff
@@ -37,73 +31,69 @@ class Listener
...
     end
 
     # Starts processing events and starts adapters
     # or resumes invoking callbacks if paused
     def start
-      transition :processing
+      transition :backend_started if state == :initializing
+      transition :frontend_ready if state == :backend_started
+      transition :processing_events if state == :frontend_ready
+      transition :processing_events if state == :paused
```
rymai: Why not use a `case` statement here?

rymai:
This PR looks definitely good. I'll try using it a bit and give you some feedback on how it works / performs for me!

Thanks for the hard work @e2!

lib/listen/listener.rb
e2: Because those steps are in a sequence - e.g. line 91 can set ':frontend_ready` and line 92 tests for that. We can't "jump" from `:processing` to `:paused`, because we have to go through the states in between (which is what's happening here). Perhaps the last line could be `if [:paused, :processing_events].include?(state)`.

rymai: Ah ok I get it, thanks for the explanation!

e2:
Thanks for finding time to go through this! There's lots more that could be done.

e2:
@rymai - if you have no objections, I'd like to release this as Listen 3.0. I want to get this version working with Spring ASAP, so people can *finally* start using Rails, recent Listen and Guard version without issues. 

While I took lots of care to make sure there are zero problems, even if I broke something, this won't affect many people, since lots of tools are locked to Listen 2.x (e.g. Guard), and if there are any issues, I prefer to discover them as soon as possible. 

So I don't see a reason for a pre-release, since for me this is master, and it's what I prefer to fix, debug and maintain - to me it's closer to "production ready" than Listen 2.x.

Let me know if you disagree.

rymai:
No problem for me. I've been using this branch this week and it worked as a charm! :+1:

Issue #335, opened 2015-08-04
optimization: check file size before checking hash


e2:
This may save hash checks on some filesystems in some cases.

Related implementation: https://github.com/guard/listen/pull/333

Also a different size is more reliable indicator of changes than atime/mtime.

ColinDKelley:
Fix merged: https://github.com/guard/listen/pull/336

PR #336, opened 2015-08-05
use file size to check for modifications

marawan31:
responds to issue #335


e2:
I'll need to rework a lot of Listen internals to make full use of this.

I'll keep this open, so it's included as soon as there's room for it in the new API.

ioquatix:
This looks good to me.

Issue #340, opened 2015-08-19
Tests fails with different orders


strzibny:
For instance they fail with `--order rand:53858` or `--order defined`:

```
+ rspec -Ilib:spec -rspec_helper spec
Run options: include {:focus=>true}
All examples were filtered out; ignoring {:focus=>true}
Randomized with seed 53858
....................................................................................................................................................................*...*.............................................................FFFFFFFFFFFFFFFFFFFFFFFFFFF..........................FATAL: Listen error: unable to monitor directories for changes.
Visit https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers for info on how to fix this.
............................
Pending: (Failures listed here are expected and do not affect your suite's status)
  1) Listen::Adapter::Config#initialize with directories as messy array implement me
     # Not yet implemented
     # ./spec/lib/listen/adapter/config_spec.rb:72
  2) Listen::Adapter::Config#initialize with no directories implement me
     # Not yet implemented
     # ./spec/lib/listen/adapter/config_spec.rb:76
Failures:
  1) Listen with one listen dir with change block raising warns the backtrace
     Got 0 failures and 2 other errors:
     1.1) Failure/Error: before { wrapper.listener.start }
          LoadError:
            cannot load such file -- rb-inotify
          # ./lib/listen/adapter/linux.rb:30:in `require'
          # ./lib/listen/adapter/linux.rb:30:in `_configure'
          # ./lib/listen/adapter/base.rb:45:in `block in configure'
          # ./lib/listen/adapter/base.rb:40:in `each'
          # ./lib/listen/adapter/base.rb:40:in `configure'
          # ./lib/listen/adapter/base.rb:63:in `start'
          # ./lib/listen/backend.rb:26:in `start'
          # ./lib/listen/listener.rb:67:in `block in <class:Listener>'
          # ./lib/listen/fsm.rb:120:in `instance_eval'
          # ./lib/listen/fsm.rb:120:in `call'
          # ./lib/listen/fsm.rb:91:in `transition_with_callbacks!'
          # ./lib/listen/fsm.rb:57:in `transition'
          # ./lib/listen/listener.rb:90:in `start'
          # ./spec/acceptance/listen_spec.rb:10:in `block (2 levels) in <top (required)>'
          # ./spec/acceptance/listen_spec.rb:17:in `block (4 levels) in <top (required)>'
          # ./spec/support/fixtures_helper.rb:20:in `fixtures'
[...]

PR #341, opened 2015-08-19
Avoid Kernel.require

strzibny:
...so we can load rb-inotify without explicit RUBYLIB path.

This fixes https://github.com/guard/listen/issues/340 and fixes the program itself.


e2:
Thanks - I restarted the build just to make sure. I think the whole lib should be mocked just like the Darwin adapter spec does it. Oh, well.

PR #344, opened 2015-08-24
Update README.md

wjordan:
Changed around the order of some information to make it less confusing and easier to read, especially near the top.
Added extra information about symlinked-directories support to the issues section.


e2:
Thanks - this will help a lot!

PR #348, opened 2015-10-06
Listen to current directory by default if no directory provided [#311]

### LINKED ISSUE #311: Listen crashes when no dir given
This crashes:

```ruby
require 'listen'
Listen.to {|*changes| puts changes.inspect }.start
```

Instead, Listen could just use the current directory as a default.

david-ding:
This fixes #311


e2:
It's perfect! Thanks!

PR #352, opened 2015-11-04
stop sharing mutable var among threads [fix #351]

### LINKED ISSUE #351: Argument being ignored in OS X
This may be something known but did not find it.

I am working on merging the new file system monitor for app reloading in Rails. The monitor works fine on Linux, but on OS X it does nothing at all. Listen is configured properly with the list of directories to watch (a bunch of them), but the callback does not get called.

I tried to reproduce outside Rails, and it seems to be related to the number of arguments. Strange... maybe that is a red herring, but this is the dummy reproduction:

Setup directories *foo*, *bar*, and *baz*. Then try this script in their parent directory:

```ruby
require 'listen'

Listen.to('foo', 'bar', 'baz') do |*args|
  p args
end.start

sleep
```

OK, touching *foo/foo.txt* and *baz/baz.txt* triggers events, but touching *bar/bar.txt* does not.

Tested with 3.0.3.

e2:
Since the local variable 'worker' was modified on each iteration, some
threads "caught" the value in the wrong state.

Using an array with an enumerator guarantees each thread receives a
block with a separate copy of the variable.

### DIFF
--- a/lib/listen/adapter/darwin.rb
+++ b/lib/listen/adapter/darwin.rb
@@ -16,32 +16,27 @@ module Listen
       # NOTE: each directory gets a DIFFERENT callback!
       def _configure(dir, &callback)
         require 'rb-fsevent'
+
         opts = { latency: options.latency }
 
         @workers ||= ::Queue.new
         @workers << FSEvent.new.tap do |worker|
+          _log :debug, "fsevent: watching: #{dir.to_s.inspect}"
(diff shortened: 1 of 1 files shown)


unjello:
This change introduced a bug in darwin adapter `_log_exception 'fsevent: running worker failed: %s: %s'` has incorrect number of parameters.

PR #357, opened 2015-11-13
do not treat non-existent paths as dirs to prevent further scanning [fix #354]

### LINKED ISSUE #354: Directory scan sometimes crashes
It looks like once in a while instead of filename as a key, and options as values sometimes values hold not a Hash but Float or Fixnum or String. I have temporarily added `values.is_a? Hash` if to the each block and it works, but didn't have time to root cause it why the values get passed like that.

Crash stack:
>E, [2015-11-12T23:32:00.713023 #62036] ERROR -- : fsevent: running worker failed: undefined method `key?' for 1447367514.0:Float: /Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/record.rb:59:in `block in dir_entries'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/record.rb:54:in `each'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/record.rb:54:in `dir_entries'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/directory.rb:10:in `scan'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/change.rb:56:in `invalidate'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/directory.rb:62:in `_change'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/directory.rb:27:in `block in scan'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/2.2.0/set.rb:283:in `each_key'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/2.2.0/set.rb:283:in `each'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/directory.rb:24:in `scan'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/change.rb:56:in `invalidate'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/adapter/base.rb:107:in `_queue_change'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/adapter/darwin.rb:45:in `block in _process_event'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/adapter/darwin.rb:40:in `each'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/adapter/darwin.rb:40:in `_process_event'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/adapter/base.rb:42:in `block (2 levels) in configure'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/rb-fsevent-0.9.6/lib/rb-fsevent/fsevent.rb:45:in `call'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/rb-fsevent-0.9.6/lib/rb-fsevent/fsevent.rb:45:in `run'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/adapter/darwin.rb:51:in `_run_worker'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/adapter/darwin.rb:35:in `_run'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/adapter/base.rb:78:in `block in start'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/internals/thread_pool.rb:6:in `call'
>/Users/angelo/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/listen-3.0.4/lib/listen/internals/thread_pool.rb:6:in `block in add'

unjello:


e2:
I think I'd prefer a `:none` result here. (Especially if there's no test for it).

Even if it increases the branch complexity, at least it's more semantically correct and more intuitive to deal with later.

unjello:
Agreed. Although to be honest it deserves a rewrite anyway. At least a minor one where you could infer the type from inspecting `previous` records. But that may be the next step. For starters - lets stop crashing :)

e2:
> Although to be honest it deserves a rewrite anyway.

The problem is finding time to do so. Listen 3.x was already a major rewrite (which happened not long ago).

> At least a minor one where you could infer the type from inspecting previous records. 

No, that approach actually failed. It's best to *never* assume what's on the filesystem may match what was recorded previously. Moving trees would just cause all kinds of problems.

> For starters - lets stop crashing :)

That's not a good approach. Every problem has to be well understood and fixed properly. Because a crash is easy to fix, but debugging a logical problem can take ages - and it's often impossible to reproduce.

Crashes are good. The solution to crashes is understanding and prevention. (Otherwise I'd be digging myself a grave...)

You use case seems strange. Let's look at the code here (simplified for clarity):

```ruby
      current = Set.new(Pathname('foo').children)
      current.each do |full_path|
        ::File.lstat(full_path.to_s)
      end
```

An `Errno::ENOENT` here suggest that the list of files (`children`) is *no longer valid*.

To me the solution is to handle `Errno::ENOENT` by rescanning the directory (because the old list obviously no longer reflects what's on the filesystem).

So I'd suggest something like this:

```ruby
begin
  current = Set.new(Pathname('foo').children)
  current.each do |full_path|
    ::File.lstat(full_path.to_s)
    # (...)
  end
rescue Errno::ENOENT
  retry
end
```

(And so `self.detect_type` shouldn't handle the exception).

What do you think?

e2:
Let me know if this solves the crashing problem: https://github.com/guard/listen/pull/358

unjello:
Fixed by #358 and released in 3.0.5 with a different fix. Thanks :)

PR #370, opened 2016-04-22
Fall back to polling w/ warning on OS X 10.6-10.8

vaz:
If OS X >= 10.9 (Darwin >= 13.0): all is well.

Else if OS X >= 10.6 (Darwin >= 10.0): if `rb-fsevent` > 0.9.4, fall back to polling and show warning recommending to use `rb-fsevent=0.9.4`

(Darwin < 10 not supported.)

See this gist for test: https://gist.github.com/vaz/fd6ee4c3563425ec11c24b96cc9da13b

### DIFF
--- a/lib/listen/adapter/darwin.rb
+++ b/lib/listen/adapter/darwin.rb
@@ -6,17 +6,34 @@ module Listen
     # Adapter implementation for Mac OS X `FSEvents`.
     #
     class Darwin < Base
-      OS_REGEXP = /darwin(1.+)?$/i
+      OS_REGEXP = /darwin(?<major_version>1\d+)/i
 
       # The default delay between checking for changes.
       DEFAULTS = { latency: 0.1 }
+
+      INCOMPATIBLE_GEM_VERSION = <<-EOS.gsub(/^ {8}/, '')
+        rb-fsevent > 0.9.4 no longer supports OS X 10.6 through 10.8.
+
+        Please add the following to your Gemfile to avoid polling for changes:
+          require 'rbconfig'
+          if RbConfig::CONFIG['target_os'] =~ /darwin(1[0-3])/i
+            gem 'rb-fsevent', '<= 0.9.4'
+          end
+      EOS
+
+      def self.usable?
+        require 'rb-fsevent'
+        darwin_version = RbConfig::CONFIG['target_os'][OS_REGEXP, :major_version] or return false
+        return true if darwin_version.to_i >= 13 # darwin13 is OS X 10.9
+        return true if Gem::Version.new(FSEvent::VERSION) <= Gem::Version.new('0.9.4')
+        Kernel.warn INCOMPATIBLE_GEM_VERSION
+        false
+      end
 
       private
 
       # NOTE: each directory gets a DIFFERENT callback!
       def _configure(dir, &callback)
-        require 'rb-fsevent'
-
         opts = { latency: options.latency }
 
         @workers ||= ::Queue.new
--- a/spec/lib/listen/adapter/darwin_spec.rb
+++ b/spec/lib/listen/adapter/darwin_spec.rb
@@ -9,9 +9,30 @@ RSpec.describe Adapter::Darwin do
   describe 'class' do
     subject { described_class }
 
-    if darwin?
+    context "on darwin 13.0 (OS X Mavericks)" do
+      before { allow(RbConfig::CONFIG).to receive(:[]).and_return('darwin13.0') }
       it { should be_usable }
-    else
+    end
+
+    context "on darwin10.0 (OS X Snow Leopard)" do
+      before { allow(RbConfig::CONFIG).to receive(:[]).and_return('darwin10.0') }
+
+      context "with rb-fsevent > 0.9.4" do
+        before { stub_const("FSEvent::VERSION", "0.9.6") }
+        it "shows a warning and should not be usable" do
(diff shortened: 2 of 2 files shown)


spec/acceptance/listen_spec.rb
```diff
@@ -7,6 +7,15 @@
...
   let(:wrapper) { setup_listener(all_options, :track_changes) }
+  around do |example|
+    warn_level = $VERBOSE
+    begin
+      $VERBOSE = nil
+      example.run
+    ensure
+      $VERBOSE = warn_level
+    end
+  end
```
vaz: I find this a bit ugly but... otherwise on OS X 10.6 to 10.8 the spec output is full of warnings. Maybe there's a nicer way, like defaulting to calling rspec with RUBYOPT='-W0'... not sure if there are warnings you'd rather not suppress though; input appreciated.

e2: What kinds of warnings? I don't see any in the Travis build log (from the earlier commit of yours).

If you really need this, I think it's best to mark which specs need this work around and use RSpec metadata to set `$VERBOSE` only on those tests. (see: https://www.relishapp.com/rspec/rspec-core/docs/metadata/user-defined-metadata).

You may also only set `$VERBOSE` for OSX at those versions.

I don't mind it being ugly - just it's not useful to make it global and on every OS the tests are run. Especially since I don't know which warnings and which specs need this.

lib/listen/adapter/darwin.rb
```diff
@@ -6,17 +6,43 @@ module Adapter
...
+        match = OS_REGEXP.match(RbConfig::CONFIG['target_os']) or return false
+        darwin_version = match[1].to_i
+        return true if darwin_version >= 13  # darwin13 is OS X 10.9 Mavericks
+
+        # OS X 10.6 through 10.8 only work with an older rb-fsevent version
+        begin
+          require 'rb-fsevent'
+          fsevent_version = Gem::Version.new(FSEvent::VERSION)
+          fsevent0_9_4 = Gem::Version.new('0.9.4') # supports 10.6-10.8
+          raise 'version too new' if fsevent_version > fsevent0_9_4
```
e2: Instead of raising and catching, you can just return true here and show the version error only to match the case, e.g.

```ruby
return true if fsevent_version <= fsevent0_9_4
Kernel.warn BUNDLER_DECLARE_GEM
false
rescue LoadError
false
end
```

(That's because a LoadError shouldn't talk about version problems, or that will confuse users).

e2: I'd also rename BUNDLER_DECLARE_GEM with e.g. UNSUPPORTED_OSX_VERSION or something like that. It's just an error message with a hint, so it should be named after the error, not the hint.

e2:
I can merge and release this after the minor changes above. 

Everything else looks good to me. 

Surprisingly lots of work was need for this.

lib/listen/adapter/darwin.rb
e2: Just as a shortcut, you can use String's built-in regexp capture:

```ruby
return false unless (darwin_major_version_string = RbConfig::CONFIG['target_os'][OS_REGEXP, 1])
return true if Integer(darwin_major_version_string) >= 13
```

Or

```ruby
darwin_version = Integer(RbConfig::CONFIG['target_os'][OS_REGEXP, 1] || "-1")
return false if darwin_version == -1
return true if darwin_version >= 13
```

e2: Just so it's clear, you don't need to overried the BUNDLER_DECLARE_GEM constant at all, because a LoadError won't (shouldn't?) happen since rb-fsevent is always a dependency (even on Linux - it just does nothing). Only BSD and Windows users really need BUNDLER_DECLARE_GEM in their adapters.

So in this case, UNSUPPORTED_OSX_VERSION would be an entirely different constant - unrelated to BUNDLER_DECLARE_GEM, even if it seems similar. (In this situation it's about fixing a broken gem vs not having an optimized backend adapter for a given os).

spec/acceptance/listen_spec.rb
vaz: The warnings are from the `Kernel.warn` that I added specifically for older Mac versions... so they would only show up in that case.

I was following the precedent from the BSD adapter in using `Kernel.warn` for that message rather than `Listen::Logger`.

On thinking it over I think it would be better to just take this out and the few contributors who might notice these warnings can just turn verbosity down from the command line if they really want to (`RUBYOPT='-W0' bundle exec rspec`). Thanks for the link on metadata though, good to know.

lib/listen/adapter/darwin.rb
vaz: Right, I see what you mean, rb-fsevent will always be present, so a LoadError should not happen. I used the same name as the BSD adapter for the warning message constant since it seemed similar (add this to your gemfile or you'll be polling) but I'll make it more descriptive.

I'm simplifying `Darwin.usable?` quite a bit, thanks for the suggestions.

e2:
It's awesome work!

For readability it might be better to implicitly return false instead of relying on `Kernel.warn` to return `nil`. But I don't mind personally since that's covered by tests.

Just squash the changes (so the history isn't as intimidating for future contributors) and I'll merge and release immediately.

vaz:
Thanks! Agreed on the explicit `false` return value, it is more readable. Added that and squashed.

Issue #392, opened 2016-05-12
Not working on OS X 10.6-10.8 due to rb-fsevent dependency change in v3.1.3


robmckinnon:
The `v3.1.3` release has removed support for OS X 10.6 through 10.8.

Commit https://github.com/guard/listen/commit/0285e25f1c7612d26f9781d4403f45cf8bbba455 changed the `rb-fsevent` dependency from:
```rb
s.add_dependency 'rb-fsevent', '>= 0.9.3'
```
to
```rb
s.add_dependency 'rb-fsevent', '~> 0.9', '>= 0.9.7'
```

`rb-fsevent > 0.9.4` no longer supports OS X 10.6 through 10.8.

PR #396, opened 2016-05-18
Backport various bugfixes up to Listen 3.1.5

e2:
Bugfixes:

- prevent crash when stopping uninitialized listener
- prevent crash when stopping uninitialized adapter
- #392 - Change rb-fsevent dependency to work on OS X 10.6-10.8
- get JRuby to fail with ENOTDIR like MRI does
- avoid unkillable thread problem on JRuby
- explicity require forwardable

Development:

- use OSX 10.10 on travis for precompiled Ruby 2.2.4
- fix broken spec (let block name conflict)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ matrix:
 os:
   - linux
   - osx
+osx_image: xcode7.1
 env:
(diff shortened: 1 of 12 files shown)


e2:
@chriseppstein - FYI - these have been released as Listen 3.0.8, which you may want to upgrade to.

PR #402, opened 2016-06-05
Adjusted acceptance_helper.rb so that the tests will pass on RBX

thedrow:


e2:
I'd instead use:

```ruby
raise unless /can't modify frozen(?: instance of)? Hash/ =~ e.message
```

Also, switch to RBX-3 in `.travis.yml`. You can remove RBX-2 completely.

e2:
Just so you know: this likely won't get the tests to pass, but we'll know what the difference in lag time is. If it's too big, it means something is broken. If it's small, it could just be that Travis is doing too much work (e.g. from having too many Listen builds running simultaneously and running out of threads and/or disk access bottlenecks).

Have you tried running the tests locally?

thedrow:
The tests pass locally with this change. At least now...

e2:
> The tests pass locally with this change. At least now...

Then it was probably just Travis being fast enough.

The lag value is configurable: https://github.com/guard/listen/blob/master/.travis.yml#L28

Ideally, it's best to know what the lag on Travis is first, before increasing it.

rymai:
Tests now pass on RBX 3 so I don't think this PR is needed anymore?

junaruga:
What we can is 
1. Do not use `coveralls` for rbx test case.
2. Adapt this PR.

junaruga:
My pull-request to RBX to fix the difference of the message was merged to master branch.
https://github.com/rubinius/rubinius/commit/91172ff60d2c8b31698d7a956d385c41bd66a9d8

spec/support/acceptance_helper.rb
```diff
@@ -218,7 +218,7 @@ def _add_changes(type, changes, dst)
     dst[type].sort!
 
   rescue RuntimeError => e
-    raise unless e.message == "can't modify frozen Hash"
+    raise unless /can't modify frozen(?: instance of)? Hash/ =~ e.message
```
rymai: I'm not sure why we need this in the first place? It seems we `.dup` each `change` in `#_relative_path`: `unfrozen_copy = change.dup` so that case shouldn't actually happen?

ioquatix:
Please rebase and resubmit the PR if it's still relevant.

thedrow:
> Please rebase and resubmit the PR if it's still relevant.

I no longer use Ruby.

PR #407, opened 2016-08-05
Pin ruby_dep to ~> 1.2.0

dennisvink:
The pin on ~> 1.2 resolves ruby_dep to the major version update released on august 3 (1.4.0), which has a dependency on ruby 2.2.5. This pull request fixes that.

### DIFF
--- a/listen.gemspec
+++ b/listen.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'rb-inotify', '~> 0.9', '>= 0.9.7'
 
   # Used to show warnings at runtime
-  s.add_dependency 'ruby_dep', '~> 1.2'
+  s.add_dependency 'ruby_dep', '~> 1.2.0'
 
   s.add_development_dependency 'bundler', '~> 1.12'
 end


e2:
Thanks, @dennisvink - what's the reason?

Generally, the dependencies here should be open so it's up to you to decide which version of Listen you want.

If you don't like the change of dependency on Ruby, it's up to you to lock to a version of Listen that depends on the Ruby you're stuck with.

I can relax the Ruby dependency in RubyDep - I just need a strong reason.

dennisvink:
Hi @e2,

Two days ago ruby_dep had a major version update moving from 1.3.x to 1.4.0, which automatically mandates Ruby 2.2.5. You pinned ruby_dep on ~> 1.2 instead of ~> 1.2.0, so you're automagically resolving to the latest ruby_dep 1.x. In my particular case it broke our CI/CD tooling.

e2:
> You pinned ruby_dep on ~> 1.2 instead of ~> 1.2.0, so you're automagically resolving to the latest ruby_dep 1.x. In my particular case it broke our CI/CD tooling.

It's a tough subject to quickly explain, so I'll give you the gist. (Ask if you want details).

Listen needs an internal overhaul and a lot of projects already rely on it a lot. So to reduce the maintenance burden for this to happen - and to keep things as stable as possible, the idea is to limit the versions that are "tested". The side-effect is breaking CI builds, but this is better than something crashing in production due to using a Ruby 2.3.x feature on Ruby 2.1.x for example.

Ideally, Ruby should be treated like any other gem dependency - it should be upgradable with a simple command, just like gems are with `gem update`.

So it's a better investment to make upgrading Ruby easier (like gems) than to have people "lock" to a given (older) Ruby version in CI and force every gem out there to support it.

Ideally Bundler should gracefully select an older version of Listen for you. This partially works on the latest versions of Bundler, but ideally Bundler should be made to resolve these for you. (So your CI builds will fallback to a earlier version of Listen on an older Ruby).

Overall it's better to have a failed CI than a crash in production. Even though this annoys many people (builds failing due to gem installation issues - due to Bundler not resolving it automagically).

Anyway, I created a Wiki page here: https://github.com/guard/listen/wiki/Ruby-version-requirements

There's a lot of confusion, so if you have any ideas how to make that wiki page clearer and more useful, feel free to do so.

Thanks for the PR anyway - to me it's a contribution already. Let me know if you need anything else.

dennisvink:
Thank you for your elaborate reply. I've pinned listen to a previous version for now.

PR #413, opened 2016-10-02
Please the 👀 with retina friendly badges

elia:


e2:
Much better. No more "castle walls" and 80s-style fonts.

Thanks!

Issue #416, opened 2016-11-17
Make rb-inotify dependency optional or provide another gem without it


jirutka:
rb-inotify depends on [ffi](https://github.com/ffi/ffi), which is very problematic:

* requires _libffi_ with its header files to be installed on the system,
* doesn’t work on hardened (Grsecurity) Linux kernel without disabling memory protection for the `ruby` binary,
* can’t be installed in Linux subsystem on Windows 10 (see https://github.com/Microsoft/BashOnWindows/issues/922).

The last one is the reason why I’m opening this issue. I’m working on a documentation site built on Middleman that depends on _listen_. Unfortunately I’m forced to support even Windows. Currently it can’t be simply installed on it, just because of a gem in the dependency tree that is not even used on this platform.

As I know, bundler doesn’t allow to exclude specific transitive dependencies. So the only option would be to fork listen, remove rb-inotify from gemspec, and point bundler to the fork’s repository.

I understand that there were some complications with conditional platform-specific dependencies. Is it still a problem? If so, what about providing another gem just without platform-specific dependencies so we can use it in Gemfile?

Related to: #54, #81, #83 (note: I didn’t read them all)

ColinDKelley:
This appears to have been fixed with the above PR.

PR #417, opened 2016-11-18
Don't require rb-fsevent until actually needed

jirutka:
rb-fsevent is used only on macOS. If you somehow exclude it from
dependencies on other platforms, for example to avoid unnecessary
complications with building native extensions, then "listen" fails to
start just because it's required before checked that it's not usable on
the current platform. This patch fixes this error.

See also #416

**NOTE:** Travis build does _not_ fail because of this change.


rymai:
Thanks @jirutka, LGTM!

PR #419, opened 2016-11-21
Require rb-fsevent when needed, in _configure method

jirutka:
This fixes an error I’ve introduced in #417, sorry for that.

The problem is: when the OS version is >= 13, then it early returns from the `usable?` function and does not require `rb-fsevent`.

```rb
      def self.usable?
        version = RbConfig::CONFIG['target_os'][OS_REGEXP, :major_version]
        return false unless version
        return true if version.to_i >= 13
        # ^--- HERE

        require 'rb-fsevent'
        ...
```


rymai:
Thanks @jirutka!

PR #401, opened 2016-06-05
Add rbx-3 to the build matrix

thedrow:

PR #427, opened 2017-04-21
Update .travis.yml

junaruga:
* I updated Rubies to latest version.
* I also added `ruby-head` as `allow_failures`.

I think adding `ruby-head` is useful. 
Because we can prepare before next version Ruby 2.5 release.
Though it might not be useful until the actual preview release.
We can support the next version as faster.

We can see this kind of logic in `rails`, `rspec` and `cucumber` and etc.
https://github.com/rails/rails/blob/master/.travis.yml
https://github.com/rspec/rspec-core/blob/master/.travis.yml
https://github.com/cucumber/cucumber-ruby/blob/master/.travis.yml

Is it possible to merge?

Thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ bundle
 .rvmrc
 .vagrant
 Gemfile.lock
+Gemfile.ruby_dep.lock
 spec/.fixtures
 coverage
 .ruby-version
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,22 +3,30 @@ bundler_args: --without development
 before_install:
   - gem install bundler
   - gem update bundler
-  - gem install ruby_dep
-  - gem update ruby_dep
+  - bundle install --gemfile=Gemfile.ruby_dep --path vendor/bundle
 rvm:
-  - 2.2.5
+  - 2.2.7
+  - 2.3.4
+  - 2.4.1
+  - ruby-head
   - jruby-head
   - rbx-2
   - jruby-9.1.2.0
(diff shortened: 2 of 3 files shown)


junaruga:
Travis all tests were failed.
Because they changed Bundler's default behavior to run with `--path` recently maybe.
For the reason, `ruby_dep` installed at `before_install` is not loaded.

https://travis-ci.org/guard/listen/builds/224367627

```
$ bundle install --without development --path=${BUNDLE_PATH:-vendor/bundle}
```

junaruga:
I changed my code. and the situation was improved.

https://travis-ci.org/guard/listen/builds/224379269

Below tests are failed because of the ruby does not exists in the environment.
Ruby 2.3.4 at Mac, Ruby 2.4.1 at Mac, rbx2 at Linux at allows_failure 

> Requested binary installation but no rubies are available to download, consider skipping --binary flag.
> Gemset '' does not exist, 'rvm rbx-2 do rvm gemset create ' first, or append '--create'.

jruby-head at Linux also failed for the test failure.

junaruga:
Finally all the test was passed.

junaruga:
After merging this PR, we must be able to merge #401 too

junaruga:
Hello, ping.

rymai:
@junaruga Thanks, LGTM!

junaruga:
@rymai thanks for the merging!

rymai:
@junaruga Thank you!

PR #430, opened 2017-06-25
Fix small markdown syntax issue in README

sivaraam:


rymai:
@sivaraam Thanks, looks good to me! :heart:

sivaraam:
@rymai You're welcome :handshake:

PR #442, opened 2018-02-02
Attempt at fixing TravisCI build

grzuy:
It appears to be giving problems in TravisCI build due to the gem being outdated in terms of the supported ruby version listed there.
It doesn't seem that's been actively updated/maintained lately either.


grzuy:
There's still an issue with jruby+linux build which i am trying to fix...

PR #443, opened 2018-03-03
Upgrade to latest rb-fsevent and rb-notify releases

thibaudgg:
[Latest rb-fsevent release](https://github.com/thibaudgg/rb-fsevent/releases/tag/0.10.3) is fixing an annoying issue, I think it would worth forcing the upgrade here.

### DIFF
--- a/listen.gemspec
+++ b/listen.gemspec
@@ -30,8 +30,8 @@ Gem::Specification.new do |s|
     abort "Install 'ruby_dep' gem before building this gem"
   end
 
-  s.add_dependency 'rb-fsevent', '~> 0.9', '>= 0.9.4'
-  s.add_dependency 'rb-inotify', '~> 0.9', '>= 0.9.7'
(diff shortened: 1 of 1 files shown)


rymai:
CI doesn't seem to like the change, though... :/

thibaudgg:
Yeah, Travis config seems to be broken here (see https://github.com/guard/listen/pull/442), @e2 could you please have a look, maybe `ruby_dep` needs an update?

junaruga:
@e2 I think at least this PR improves current situation.

Current master
https://travis-ci.org/guard/listen/builds/248699966

It is better to merge isn't it?
Then we can change `rbx-2` to `rbx-3`, and add Ruby 2.5 in other PRs, and check jruby issue etc.

rymai:
@junaruga Sounds good!

junaruga:
@rymai thanks for merging!

PR #454, opened 2018-09-04
Locked platform dependencies inside case block in gemspec

majioa:


listen.gemspec
```diff
@@ -24,8 +24,12 @@ Gem::Specification.new do |s|
...
   s.required_ruby_version = ['~> 2.2', '>= 2.2.7']
 
-  s.add_dependency 'rb-fsevent', '~> 0.10', '>= 0.10.3'
-  s.add_dependency 'rb-inotify', '~> 0.9', '>= 0.9.10'
+  case RUBY_PLATFORM
+  when /darwin/i
+    s.add_dependency 'rb-fsevent', '~> 0.10', '>= 0.10.3'
+  when /linux/i
+    s.add_dependency 'rb-inotify', '~> 0.9', '>= 0.9.10'
+  end
```
rymai: I don't think this is going to work since `.gemspec` are evaluated at the time the gem is build (i.e. only the dependencies matching the machine that builds it will be listed). See https://stackoverflow.com/a/41694151/304055.

majioa: I read it but requirements on building a gem depending on the ENV variable is not the same case, because the dependency here is on platform i.e. exactly build/install env.

rymai: I think the problem is the same, the `.gemspec` is evaluated at build time, thus one the developer's machine. For instance, if I were to release a new version, the built gem would depend on `rb-fsevent` only since I'm on macOS, and people on Linux would need to manually install `rb-inotify`.

majioa: may be it should be ok, and just be explicitly described in manual

rymai: Hmm, I think that was discussed before and we decided that the best trade-off was to keep both dependencies in the `.gemspec`. I'm going to close this PR since we cannot accept it. Thanks for contributing!

PR #459, opened 2019-03-01
Dir is more than normal , try multi thread handle

WoolenWang:
When there is many dir try multi thread handle this will make it more quick, it fix my problem.

### DIFF
--- a/lib/listen/record.rb
+++ b/lib/listen/record.rb
@@ -103,13 +103,13 @@ module Listen
         thr_arr << Thread.new(remaining_, symlink_detector_) do |remaining, symlink_detector|
           begin
             entry = remaining.pop(true)
-          rescue Exception
-            return
+          rescue StandardError
+            return nil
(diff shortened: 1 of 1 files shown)


lib/listen/record.rb
```diff
@@ -97,16 +97,28 @@ def _fast_unset_path(dirname, basename)
...
-    rescue Errno::ENOTDIR
-      _fast_try_file(entry)
-    rescue SystemCallError, SymlinkDetector::Error
-      _fast_unset_path(entry.relative, entry.name)
+    def _fast_build_dir(remaining_, symlink_detector_)
+      thr_arr = []
+      5.times do
+        thr_arr << Thread.new(remaining_, symlink_detector_) do |remaining, symlink_detector|
+          begin
+            entry = remaining.pop(true)
```
ioquatix: It's not thread safe.

PR #460, opened 2019-03-12
Fix: Tracking of removed subdirectories

bryanlira:
Some MacOS commands weren't recognized by the listener. example: 
- ⌘command + delete 
- Retrieve directories from the recycling bin


ioquatix:
Do you mind rebasing this on master.

PR #464, opened 2019-08-07
Fix warnings in spec output

floehopper:
This fixes two types of warnings which were appearing in the spec output thus reducing the signal-to-noise ratio.


rymai:
@floehopper Thank you, looks good to me! :heart: :yellow_heart: :green_heart: :rocket:

PR #469, opened 2019-10-25
Ignore mutagen sync temporary files

nilbus:
[Mutagen](https://mutagen.io) is a sync tool for remote development that copies files into place on a remote machine [running listen/guard] when edits are made on a local machine. To perform an atomic update, it creates a temp/staging file and moves it into place.

Normally this results in an addition, which results in the change being ignored by many guard plugins. Listen's `QueueOptimizer` can [detect this][] as an "editor save" when it detects that a silenced temp file is moved to replace a watched file.

This patch updates the ignore pattern to detect mutagen temp files, silencing them and allowing the staging file move to be detected properly as a modification.

[detect this]: https://github.com/guard/listen/blob/2fe70d6226805410f6ba9790eb1acca4533bcb75/lib/listen/queue_optimizer.rb#L128

Post-merge tasks
---

- [ ] Update [Guardfile DSL wiki ignore section](https://github.com/guard/guard/wiki/Guardfile-DSL---Configuring-Guard#ignore) to include this new default

### DIFF
--- a/lib/listen/silencer.rb
+++ b/lib/listen/silencer.rb
@@ -46,6 +46,9 @@ module Listen
          )
         )
 
+      # Mutagen sync temporary files
+      | \.mutagen-temporary.*
+
       # other files
       | \.DS_Store
       | \.tmp
--- a/spec/lib/listen/silencer_spec.rb
+++ b/spec/lib/listen/silencer_spec.rb
@@ -1,5 +1,3 @@
-# encoding: UTF-8
-
 RSpec::Matchers.define :accept do |type, path|
   match { |actual| !actual.silenced?(Pathname(path), type) }
 end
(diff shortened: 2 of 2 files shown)


nilbus:
Mutagen creates all its files with a `.mutagen-temporary` prefix ([reference][]). I put the prefix pattern ending in `.*` in the suffixes where all other tempfile patterns are. Matching will be as performant as the others, but let me know if you think it should go into its own separate ignore prefixes list since it's not technically a suffix.

[reference]: https://github.com/mutagen-io/mutagen/blob/090ef5f4741bdd2b1a98a1cf3f8338703483b688/pkg/filesystem/temporary.go#L12

benjaminoakes:
Thank you for addressing this

rymai:
@nilbus Thank you, looks good to me! :heart: :yellow_heart: :green_heart: :rocket:

PR #470, opened 2019-12-02
Use one `fsevent_watch` process per listener instead of one per dir

tgxworld:
Related: https://github.com/rails/rails/issues/26158, https://github.com/rails/rails/issues/37269

This is an attempt to resolve the problem of too many `fsevent_watch` events being spun up on OSX since the number of processes increases with the number of directories being watched.

Screenshot 2019-12-02 at 2 20 28 PM


tgxworld:
cc @ioquatix @rymai Can I get your thoughts regarding this before I look into adding the specs? Thank you!

ioquatix:
Yeah, we should definitely fix this.

spec/lib/listen/adapter/darwin_spec.rb
```diff
@@ -69,130 +70,4 @@
       it { should eq 1234 }
     end
   end
-
-  describe 'multiple dirs' do
```
tgxworld: Specs are no longer relevant and there is already high level acceptance test that runs for OSX on CI.

lib/listen/adapter/darwin.rb
```diff
@@ -35,59 +35,45 @@ def self.usable?
...
-      def _process_event(dir, event)
-        _log :debug, "fsevent: processing event: #{event.inspect}"
-        event.each do |path|
-          new_path = Pathname.new(path.sub(%r{\/$}, ''))
-          _log :debug, "fsevent: #{new_path}"
-          # TODO: does this preserve symlinks?
-          rel_path = new_path.relative_path_from(dir).to_s
-          _queue_change(:dir, dir, rel_path, recursive: true)
+      def _process_changes(dirs)
+        dirs.each do |dir|
```
tgxworld: I'm aware there are two loops here but for the average use case, this is going to consist of only a couple of directories that change each time.

tgxworld:
@ioquatix Would you mind reviewing this? Thank you so much! 🙇

ioquatix:
I will review it tonight.

tgxworld:
Thank you so much 👍

ioquatix:
Using local branch: https://github.com/guard/listen/pull/471

spec/lib/listen/adapter/darwin_spec.rb
ioquatix: Oh, I see. Thanks.

PR #471, opened 2019-12-05
Use one `fsevent_watch` process per listener instead of one per dir

ioquatix:


spec/lib/listen/adapter/darwin_spec.rb
```diff
@@ -69,130 +69,4 @@
       it { should eq 1234 }
     end
   end
-
-  describe 'multiple dirs' do
```
ioquatix: @tgxworld why did we have to remove all these specs?

ioquatix:
@tgxworld thanks for your work.

Issue #476, opened 2020-06-26
Memory leaks when using Listener#stop


jonathanhefner:
I've noticed at least two memory "leaks" when using on `Listener#stop` (not `Listen.stop`).

The first leak is the `@listeners` array in the `Listen` module.  It is appended to [when creating a `Listener`](https://github.com/guard/listen/blob/ba5059c9f3a8594eae7e0050687c1a6735046d85/lib/listen.rb#L37), but only cleared in [`Listen.stop`](https://github.com/guard/listen/blob/ba5059c9f3a8594eae7e0050687c1a6735046d85/lib/listen.rb#L49) (not `Listener#stop`).  I suppose a workaround would be to use `Listener.new` directly instead of `Listen.to`, but that doesn't seem to be the recommended usage (going by the README).  Also note that this leak isn't merely of `Listener` instances, but of potentially large object graphs that each `Listener` instance's block holds a reference to.

The second leak is `Listen::Internals::ThreadPool`.  It is appended to in [`Listen::Adapter::Base#start`](https://github.com/guard/listen/blob/ba5059c9f3a8594eae7e0050687c1a6735046d85/lib/listen/adapter/base.rb#L73) and [`Listen::Event::Loop#setup`](https://github.com/guard/listen/blob/ba5059c9f3a8594eae7e0050687c1a6735046d85/lib/listen/event/loop.rb#L41), but only cleared in [`Listen.stop`](https://github.com/guard/listen/blob/ba5059c9f3a8594eae7e0050687c1a6735046d85/lib/listen.rb#L44) (not `Listener#stop`).

My proposal to fix the first leak is to remove `@listeners` and replace its usage in `Listen.stop` with [`ObjectSpace.each_object`](https://docs.ruby-lang.org/en/2.2.0/ObjectSpace.html#method-c-each_object).  It will be slower than iterating an array, but it shouldn't be done often, so that shouldn't be an issue.

My proposal to fix the second leak is to remove `Listen::Internals::ThreadPool` entirely, and store thread references directly in each `Listen::Adapter::Base` and `Listen::Event::Loop` instance.  Then `Listen::Adapter::Base#stop` and `Listen::Event::Loop#teardown` can kill and clear an instance's own threads.

I am completely new to the Listen code base, so please tell me if I am misunderstanding the situation, or if my ideas are laughably naive. :sweat_smile:  However, if these proposals sound good, I am willing to submit a PR.

jonathanhefner:
[...]

PR #477, opened 2020-07-05
Wrap Listener instances in WeakRef

jonathanhefner:
The `@listeners` variable holds a reference to each `Listener` instance created by `Listen.to` and can only be cleared by calling `Listen.stop`.  This can prevent `Listener` instances from being garbage collected if they are abandoned before `Listen.stop` is called.

This commit wraps `Listener` instances in `WeakRef` before adding them to `@listeners`, to allow them to be garbage collected.

Partially addresses #476.

### DIFF
--- a/lib/listen.rb
+++ b/lib/listen.rb
@@ -1,4 +1,5 @@
 require 'logger'
+require 'weakref'
 require 'listen/logger'
 require 'listen/listener'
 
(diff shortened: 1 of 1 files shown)


jonathanhefner:
@headius  Does this look all right to you?  I decided to forgo clearing dead `WeakRef`s in `Listen.to` because it would require using `Mutex` instead of a simple `Queue` to stay thread-safe.  I can make that change if you prefer.  However, it might be better to recommend `Listener.new` over `Listen.to` when people are creating lots and lots of listeners (since `Listener.new` has less overhead).

PR #480, opened 2020-07-23
Remove Listen::Internals::ThreadPool

### LINKED ISSUE #476: Memory leaks when using Listener#stop

jonathanhefner:
`Listen::Internals::ThreadPool` manages all listener threads.  Thus the only way to kill and subsequently garbage collect listener threads is by calling the `Listen::Internals::ThreadPool.stop` method (typically done via `Listen.stop`).  This is a problem when individual listeners must be stopped and abandoned for garbage collection.

This commit removes `Listen::Internals::ThreadPool` in favor of listener instances managing their own threads.

Partially addresses #476.

---

/cc @headius

This is based on top of #477, so that should probably be merged first.  Together, these PRs fix #476.


ioquatix (approved):

ioquatix:
I am manually rebasing and merging.

Issue #481, opened 2020-08-06
Linux backend can hang at startup on sleep in Listen::Event::Loop#_wait_until_resumed


ColinDKelley:
We have been running `listen` in production for ~6 months across hundreds of k8s pods with frequent file system changes to listen for. There is a bug where sometimes the backend loop code would simply hang and never process changes. We have traced that to this code in `Listen::Event::Loop`:
```
def _wait_until_resumed(ready_queue)
  self.state = :paused
  ready_queue << :ready
  sleep
  self.state = :processing
end
```
That call to `sleep` is expecting to be awakened by this state transition:
```
state :processing_events, to: [:paused, :stopped] do
  processor.resume
end
```
which is defined as:
```
def resume
  fail Error::NotStarted if stopped?
  return unless wait_thread
  _wakeup(:resume)
end
...
def _wakeup(reason)
  @reasons << reason
  wait_thread.wakeup
end
```
The bug here is a race condition. If the `wakeup` runs before the thread calls `sleep`, there is no memory of that, so the `sleep` simply blocks forever in the `:paused` state.

I believe the proper way to fix this is to have a `::Mutex` protecting the state and associated `::Condition` object(s) to fire on state transitions that need to be waited for. There is no need to have the `::Queue` objects `ready_queue` and `@reasons` nor any of the calls to `sleep` or `Thread#wakeup`.

ColinDKelley:
Merged and ready to release in v3.3.0.

PR #482, opened 2020-08-12
Issue #481 fix hanging race condition

ColinDKelley:
Fixes issue #481 by removing code that attempted to synchronize threads with `sleep()` and `Thread#wakeup`. That was a race condition bug since if `wakeup` ran before `sleep()`, it would be lost and the sleeping thread would simply hang.

Detailed changes:

- Use :`:Mutex`/`::ConditionVariable` to wait/wake up on state transitions. I factored this down into FSM since it's used in two state machine classes now.
- Clean up and simplify `FSM`. Enforce symbol contracts, remove unnecessary methods, rename `default_state` to `start_state` since the latter is standard FSM terminology. Also initialize with `initialize_fsm` vs `super()`.
- Use `Queue#pop` to block front-end when waiting for events.
- Combine `Loop#setup` and `Loop#resume` since they were always called consecutively. This allowed the `Listener` `:front_end_ready` state to disappear.
- Remove `Loop#resume` and `Loop#pause` since they were no-ops.
- Rename `teardown` to `stop` to be consistent. (At some point I'd like to do a similar thing with `:processing_events`. This could simply be called `:started`.)

### DIFF
--- a/lib/listen/event/loop.rb
+++ b/lib/listen/event/loop.rb
@@ -14,32 +14,26 @@ module Listen
       def initialize(config)
         @config = config
         @wait_thread = nil
-        @state = :paused
+        @state = :pre_start # ... :starting, :started, :stopped
         @reasons = ::Queue.new
       end
 
       def wakeup_on_event
-        return if stopped?
-        return unless processing?
-        return unless wait_thread.alive?
-        _wakeup(:event)
+        if started? && @wait_thread.alive?
+          _wakeup(:event)
+        end
       end
 
-      def paused?
-        wait_thread && state == :paused
+      def started?
+        @state == :started
       end
 
-      def processing?
-        return false if stopped?
-        return false if paused?
-        state == :processing
-      end
-
-      def setup
+      def start
         # TODO: use a Fiber instead?
+        @state = :starting
         q = ::Queue.new
         @wait_thread = Thread.new do
-          _wait_for_changes(q, config)
+          _wait_for_changes(q)
         end
 
         Listen::Logger.debug('Waiting for processing to start...')
@@ -47,8 +41,8 @@ module Listen
       end
 
       def resume
-        fail Error::NotStarted if stopped?
-        return unless wait_thread
+        fail Error::NotStarted if @state == :pre_start
+        return unless @wait_thread
         _wakeup(:resume)
       end
 
@@ -58,30 +52,26 @@ module Listen
       end
 
       def teardown
-        return unless wait_thread
-        if wait_thread.alive?
+        return if stopped?
+        if @wait_thread.alive?
           _wakeup(:teardown)
-          wait_thread.join.kill
+          @wait_thread.join.kill
         end
         @wait_thread = nil
+        @state = :stopped
       end
 
       def stopped?
-        !wait_thread
+        @state == :stopped
       end
 
       private
 
-      attr_reader :config
-      attr_reader :wait_thread
-
-      attr_accessor :state
-
-      def _wait_for_changes(ready_queue, config)
-        processor = Event::Processor.new(config, @reasons)
+      def _wait_for_changes(ready_queue)
+        processor = Event::Processor.new(@config, @reasons)
 
         _wait_until_resumed(ready_queue)
-        processor.loop_for(config.min_delay_between_events)
+        processor.loop_for(@config.min_delay_between_events)
       rescue StandardError => ex
         _nice_error(ex)
       end
@@ -91,10 +81,9 @@ module Listen
       end
 
       def _wait_until_resumed(ready_queue)
-        self.state = :paused
         ready_queue << :ready
         sleep
-        self.state = :processing
+        @state = :started
       end
 
       def _nice_error(ex)
@@ -110,7 +99,7 @@ module Listen
 
       def _wakeup(reason)
         @reasons << reason
-        wait_thread.wakeup
+        @wait_thread.wakeup
       end
     end
   end
--- a/lib/listen/event/loop.rb
+++ b/lib/listen/event/loop.rb
@@ -60,7 +60,7 @@ module Listen
         # fail NotImplementedError
       end
 
-      def teardown
+      def stop
         return if stopped?
         transition! :stopped
 
--- a/lib/listen/event/loop.rb
+++ b/lib/listen/event/loop.rb
@@ -23,7 +23,7 @@ module Listen
         @config = config
         @wait_thread = nil
         @reasons = ::Queue.new
-        super()
+        initialize_fsm
       end
 
       def wakeup_on_event
--- a/lib/listen/event/loop.rb
+++ b/lib/listen/event/loop.rb
@@ -6,18 +6,24 @@ require 'listen/event/processor'
 module Listen
   module Event
     class Loop
+      include Listen::FSM
+
       class Error < RuntimeError
         class ThreadFailedToStart < Error; end
         class AlreadyStarted < Error; end
       end
 
+      start_state :pre_start
+      state :pre_start
+      state :starting
+      state :started
+      state :stopped
+
       def initialize(config)
         @config = config
         @wait_thread = nil
-        @mutex = ::Mutex.new
-        @state_changed = ::ConditionVariable.new
-        @state = :pre_start # ... :starting, :started, :stopped
         @reasons = ::Queue.new
+        super()
       end
 
       def wakeup_on_event
@@ -27,15 +33,15 @@ module Listen
       end
 
       def started?
-        @state == :started
+        state == :started
       end
 
       MAX_STARTUP_SECONDS = 5.0
 
       def start
-       # TODO: use a Fiber instead?
-      _transition!(:starting) do
-          @state == :pre_start or raise Error::AlreadyStarted
+        # TODO: use a Fiber instead?
+        transition! :starting do
+          state == :pre_start or raise Error::AlreadyStarted
         end
 
         @wait_thread = Thread.new do
@@ -44,7 +50,7 @@ module Listen
 
         Listen::Logger.debug("Waiting for processing to start...")
 
-        _wait_for_state(:started, MAX_STARTUP_SECONDS) or raise Error::ThreadFailedToStart, "thread didn't start in #{MAX_STARTUP_SECONDS} seconds (in state: #{@state.inspect})"
+        wait_for_state(:started, MAX_STARTUP_SECONDS) or raise Error::ThreadFailedToStart, "thread didn't start in #{MAX_STARTUP_SECONDS} seconds (in state: #{state.inspect})"
 
         Listen::Logger.debug('Processing started.')
       end
@@ -56,7 +62,7 @@ module Listen
 
       def teardown
         return if stopped?
-        _transition!(:stopped)
+        transition! :stopped
 
         if @wait_thread.alive?
           @wait_thread.join.kill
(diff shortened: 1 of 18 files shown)


lib/listen/fsm.rb
```diff
@@ -1,131 +1,128 @@
...
+        raise ArgumentError, msg
       end
 
-      new_state = states[state_name]
-
-      unless new_state
-        return if state_name == default_state
-        fail ArgumentError, "invalid state for #{self.class}: #{state_name}"
+      unless (new_state = self.class.states[new_state_name])
+        new_state_name == self.class.start_state or raise ArgumentError, "invalid state for #{self.class}: #{new_state_name}"
```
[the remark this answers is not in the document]
ColinDKelley: No.

lib/listen/fsm.rb
```diff
@@ -1,131 +1,128 @@
...
+      end
     end
 
-    protected
+    # checks for one of the given states to wait for
+    # if not already, waits for a state change (up to timeout seconds--`nil` means infinite)
+    # returns truthy iff the transition to one of the desired state has occurred
+    def wait_for_state(*wait_for_states, timeout: nil)
+      @mutex.synchronize do
+        if !wait_for_states.include?(@state)
```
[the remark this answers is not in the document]
ColinDKelley: No thanks.

lib/listen/fsm.rb
```diff
@@ -1,131 +1,128 @@
...
+      end
     end
 
-    # Immediate state transition with no checks, or callbacks. "Dangerous!"
-    def transition!(state_name)
-      @state = state_name
+    # Low-level, immediate state transition with no checks or callbacks.
+    def transition!(new_state_name)
+      new_state_name.is_a?(Symbol) or raise ArgumentError, "state name must be a Symbol (got #{new_state_name.inspect})"
+      @mutex or raise ArgumentError, "FSM not initialized. You must call super() from initialize!"
```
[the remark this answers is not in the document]
ColinDKelley: No and no.

lib/listen/fsm.rb
```diff
@@ -1,131 +1,128 @@
...
+        transition_with_callbacks!(new_state)
+      end
     end
 
-    # Immediate state transition with no checks, or callbacks. "Dangerous!"
-    def transition!(state_name)
-      @state = state_name
+    # Low-level, immediate state transition with no checks or callbacks.
+    def transition!(new_state_name)
+      new_state_name.is_a?(Symbol) or raise ArgumentError, "state name must be a Symbol (got #{new_state_name.inspect})"
```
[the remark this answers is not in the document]
ColinDKelley: No.

lib/listen/event/loop.rb
```diff
@@ -6,97 +6,86 @@
...
-          _wait_for_changes(q, config)
+          _process_changes
         end
 
-        Listen::Logger.debug('Waiting for processing to start...')
-        Timeout.timeout(5) { q.pop }
-      end
+        Listen::Logger.debug("Waiting for processing to start...")
+
+        wait_for_state(:started, MAX_STARTUP_SECONDS) or raise Error::ThreadFailedToStart, "thread didn't start in #{MAX_STARTUP_SECONDS} seconds (in state: #{state.inspect})"
```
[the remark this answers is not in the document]
ColinDKelley: No on `||`.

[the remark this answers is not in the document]
ioquatix: To make this line a bit more managable, you could write:

``` ruby
unless wait_for_state(...)
  raise Error ...
end
```

Because this is very important, I'd say it's better to have explicit lines for each part.

lib/listen/event/loop.rb
```diff
@@ -6,97 +6,86 @@
 module Listen
   module Event
     class Loop
+      include Listen::FSM
+
       class Error < RuntimeError
-        class NotStarted < Error
-        end
+        class ThreadFailedToStart < Error; end
```
ioquatix, line 12: Since `NotStarted` is a public interface, please reuse it instead of introducing `ThreadFailedToStart`.

lib/listen/event/loop.rb
```diff
@@ -6,97 +6,86 @@
...
-      def processing?
-        return false if stopped?
-        return false if paused?
-        state == :processing
-      end
+      MAX_STARTUP_SECONDS = 5.0
+
+      def start
+        transition! :starting do
+          state == :pre_start or raise Error::AlreadyStarted
```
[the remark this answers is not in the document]
ioquatix: I would say it's better to use explicit if statement here

``` ruby
unless state == :pre_start
  raise Error::AlreadyStarted...
end
```

lib/listen/event/loop.rb
```diff
@@ -6,97 +6,86 @@
 module Listen
   module Event
     class Loop
+      include Listen::FSM
+
       class Error < RuntimeError
-        class NotStarted < Error
-        end
+        class ThreadFailedToStart < Error; end
+        class AlreadyStarted < Error; end
```
ioquatix, line 13: I wonder if this should inherit from `NotStarted` to preserve existing semantics where possible.

lib/listen/fsm.rb
```diff
@@ -1,131 +1,128 @@
 # Code copied from https://github.com/celluloid/celluloid-fsm
+
+require 'thread'
+
 module Listen
   module FSM
-    DEFAULT_STATE = :default # Default state name unless one is explicitly set
+    START_STATE = :default # Start state name unless one is explicitly set
```
ioquatix, line 7: This looks okay to me but are we breaking a public interface?

lib/listen/fsm.rb
```diff
@@ -1,131 +1,128 @@
...
-      # Obtain or set the default state
-      # Passing a state name sets the default state
-      def default_state(new_default = nil)
-        if new_default
-          @default_state = new_default.to_sym
+      # Obtain or set the start state
+      # Passing a state name sets the start state
+      def start_state(new_start_state = nil)
+        if new_start_state
+          new_start_state.is_a?(Symbol) or raise ArgumentError, "state name must be a Symbol (got #{new_start_state.inspect})"
```
ioquatix, line 19: Instead of explicitly detecting it's a symbol, why not use `to_sym`?

lib/listen/fsm.rb
ioquatix, line 61: This seems like an overly specific issue to check for - I can imagine some code was not working correctly, but in that case `@mutex.synchronize` will blow up with nil error... do we need to manually check it?

lib/listen/fsm.rb
```diff
@@ -1,131 +1,128 @@
...
     class State
       attr_reader :name, :transitions
 
-      def initialize(name, transitions = nil, &block)
+      def initialize(name, transitions, &block)
         @name = name
         @block = block
-        @transitions = nil
-        @transitions = Array(transitions).map(&:to_sym) if transitions
+        @transitions = if transitions
```
ioquatix, line 114: You don't need `if transitions`.

ioquatix (changes requested):
Thanks for your effort. I have asked for some changes.

In the future, please make smaller PRs focusing on one particular issue.

Pay attention to changes to public interfaces since this can cause a lot of pain downstream.

I look forward to reviewing and merging this PR and others from you.

ioquatix:
You may need to rebase on master.

lib/listen/event/loop.rb
ioquatix: Yes, understood. In this context... if someone calls start, and it's already started, that's a kind of `NotStarted` because it was already started. If you can think of something better, it's worth considering.

lib/listen/fsm.rb
ioquatix: This kind of pseudo type checking is uncommon in Ruby, even thought I can see how it makes sense.

The reality is, if you add this, there are many other places where you might start trying to enforce type contracts.

Agreed, we don't really want users passing in strings.

lib/listen/fsm.rb
ioquatix: Understood.

Well, I don't really want to make the code any more complicated than it already is.

If the code is not implemented correctly (missing call to `super`) it's a mechanical problem - i.e. the code needs to be fixed. This isn't a transient run-time issue. The current implementation is okay, but it's also very explicit in a non-canonical way.

I don't think we should be adding extra instance variables to enforce initialisation.

lib/listen/fsm.rb
ioquatix: It's simply not needed. `Array(nil) => []`.

lib/listen/event/loop.rb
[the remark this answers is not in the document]
ioquatix: I understand where you are coming from.

For really important logic like this, I like to read the critical condition first. Code which puts the most important part last seems obscure to me.

spec/lib/listen/fsm_spec.rb
```diff
@@ -0,0 +1,103 @@
+RSpec.describe Listen::FSM do
```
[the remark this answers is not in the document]
ColinDKelley: I will make a separate PR for this later.

lib/listen/fsm.rb
ColinDKelley: I don't follow. That `if` is there so that when `transitions` is `nil`, `@transitions` will be as well. The code below counts on that, because it uses `nil` in `@transitions` as an indicator that _all_ transitions are allowed.

lib/listen/event/loop.rb
ColinDKelley: I restored it to its old name, `NotStarted`. And if `start` is called when we're already started, I put the code back as it was to simply return as a no-op, since there could be callers who were counting on that public behavior.
https://github.com/guard/listen/pull/482/commits/0b236c1ab903c8653626e8d6404fd1c680238aa8

lib/listen/event/loop.rb
ColinDKelley: Restored to its old behavior: https://github.com/guard/listen/pull/482/commits/0b236c1ab903c8653626e8d6404fd1c680238aa8

lib/listen/fsm.rb
ColinDKelley: True that it's not common in Ruby (yet). But we've found it to be invaluable for containing String vs. Symbol insanity! Otherwise, you wind up with defensive code sprinkled everywhere that tries to handle String and Symbol as equivalent...and you never know if that clutter is complete.

Speaking of which, I added the missing spec for the FSM module and made sure to test this case: https://github.com/guard/listen/pull/482/commits/bbd10eecca86cfdae4f73a3a4906d8554f086b19

lib/listen/fsm.rb
ColinDKelley: I realized we had to move away from  `super()`. There's a reason I've never seen mixins initialized that way before: it precludes any base class initialization, since the mixin would have to do that for you (by calling `super` itself), but would have no way to know the base class's `initialize` signature!

Instead I went with an explicitly-named method `initialize_fsm`. And stored a variable to remember if that didn't happen. And I addressed the longstanding comment in there to make the FSM methods private. And I added a spec for the FSM module and tested all of this. I think it's all much cleaner now! LMK what you think.

ColinDKelley:
@ioquatix I believe all comments are addressed here. I merged up latest master as well. There were a couple build failures that came across with master. I sent a separate PR https://github.com/guard/listen/pull/488 for those and merged it in here.

ColinDKelley:
Hi @ioquatix
Anything else to do on this PR?

ioquatix:
My apologies, I was busy for Ruby Kaigi conference. I'm reviewing it now.

ioquatix:
I'll fix the Ruby 3 issues. However can you check why:

```
Failures:

  1) Listen with one listen dir force_polling option to false with default ignore options two dirs with files in listen dir listens to dir move
     Failure/Error:
       expect(wrapper.listen do
         mv 'dir1', 'dir2/'
       end).to eq expected

       expected: {:added=>["dir2/dir1/file1.rb"], :modified=>[], :removed=>["dir1/file1.rb"]}
            got: {:added=>["dir2/dir1/file1.rb"], :modified=>["dir2/file2.rb"], :removed=>["dir1/file1.rb"]}

       (compared using ==)
```

ioquatix:
Okay, I see a bunch of specs are failing, I'll try to bring master back to green.

ioquatix:
Okay, master is green again, can you rebase?

ColinDKelley:
@ioquatix Thanks for getting `master` green again. I've merged this branch up to it.
I will look for that failure of

> Listen with one listen dir force_polling option to false with default ignore options two dirs with files in listen dir listens to dir move

ioquatix:
Can you please rebase on master so I can merge - please don't merge master into your branch because it creates a huge mess.

image

ColinDKelley:
@ioquatix : Rebased and pushed.

ioquatix:
Great work, thanks for your effort. Do you have time to jump on call some time this week to discuss a release plan/schedule?

ColinDKelley:
Hi @ioquatix Thank you! Yes, this week is good for a call. I'm on `US/Pacific` time. How about if we pick an exact time and medium by email?

PR #488, opened 2020-08-27
fix file_spec by stubbing size: 1

ColinDKelley:
I noticed that travis-ci tests started failing with https://github.com/guard/listen/commit/debf5cbec3fbba7c0ea9f05f9868af2412b4a828 .


ioquatix:
Oh, thanks for this, I didn't even see it, my bad.

Issue #487, opened 2020-08-27
move thread caller stack and rescue+log to a common place


ColinDKelley:
I was happy to see this PR remove the "ThreadPool": https://github.com/guard/listen/pull/483 . It was a nice improvement to track the threads inside the listeners rather than globally.

All the threads should have a rescue-and-log at the top. And it's very handing to remember the original caller who created the thread to help identify it. That's done here:
https://github.com/guard/listen/blob/master/lib/listen/adapter/base.rb#L72-L85

Let's put that in a general place and use it everywhere that threads are created. (And not `raise` from the top of the thread. That falls into some odd behavior in Ruby, where that exception can later be raised by code that calls `join`.)

Also: I don't believe the `.kill` is necessary here, since once a `Thread#join` returns, that means the thread has exited, right?
https://github.com/guard/listen/blob/master/lib/listen/event/loop.rb#L64

ColinDKelley:
@jonathanhefner Are you good with the changes I suggested here? If so, I'll make a PR.
1. Creating a wrapper around thread creation that keeps track of the thread creator and rescues/logs any exceptions, including that info.
2. Dropping `.kill` after `thread.join`. The Ruby [Thread#join documentation](https://ruby-doc.org/core-2.7.0/Thread.html#method-i-join) says:

> Does not return until thr exits or until the given limit seconds have passed.

And we're not passing `limit` seconds, so the part after `or` doesn't apply here.

jonathanhefner:
@ColinDKelley  I'm not actually affiliated with the Listen project, so I may not be the right person to ask.  (I was fixing #476 to address some downstream issues in Rails.)

> 1. Creating a wrapper around thread creation that keeps track of the thread creator and rescues/logs any exceptions, including that info.

It might be good to ellaborate more about how this would be useful.  From a cursory search, it looks like there are three places where `Thread.new` is called, and each has a `rescue`:

* `Listen::Adapter::Base` (the one you mentioned): [`Thread.new`](https://github.com/guard/listen/blob/86e06aeb0878dcd46fd635d620656748255d3f67/lib/listen/adapter/base.rb#L73) / [`rescue`](https://github.com/guard/listen/blob/86e06aeb0878dcd46fd635d620656748255d3f67/lib/listen/adapter/base.rb#L79)
[...]

PR #496, opened 2020-09-20
issue #487: unify Thread.new

ColinDKelley:
Issue #487:
- Unify thread creation through new `Listen::Thread` module.
- Set the `Thread#name =` property to assist with debugging in the surrounding process.
- `rescue` all exceptions in `Listen::Thread` and log them, including complete call stack (caller + exception backtrace) and show `cause` detail if any.
- Only call one of `Thread#join`, `Thread.kill`.

### DIFF
--- a/lib/listen/adapter/darwin.rb
+++ b/lib/listen/adapter/darwin.rb
@@ -68,14 +68,6 @@ module Listen
         _queue_change(:dir, dir, rel_path, recursive: true)
       end
 
-      def _run_worker(worker)
-        Listen.logger.debug { "fsevent: running worker: #{worker.inspect}" }
-        worker.run
-      rescue
-        format_string = 'fsevent: running worker failed: %s:%s called from: %s'
-        _log_exception format_string, caller
-      end
-
       def _stop
         @worker_thread&.kill
         super
--- a/lib/listen/adapter/darwin.rb
+++ b/lib/listen/adapter/darwin.rb
@@ -44,7 +44,7 @@ module Listen
         require 'rb-fsevent'
         worker = FSEvent.new
         dirs_to_watch = @callbacks.keys.map(&:to_s)
-        _log(:info) { "fsevent: watching: #{dirs_to_watch.inspect}" }
+        Listen.logger.info { "fsevent: watching: #{dirs_to_watch.inspect}" }
         worker.watch(dirs_to_watch, { latency: options.latency }, &method(:_process_changes))
         @worker_thread = Thread.new { _run_worker(worker) }
       end
@@ -62,14 +62,14 @@ module Listen
       end
 
       def _process_event(dir, path)
-        _log(:debug) { "fsevent: processing path: #{path.inspect}" }
+        Listen.logger.debug { "fsevent: processing path: #{path.inspect}" }
         # TODO: does this preserve symlinks?
         rel_path = path.relative_path_from(dir).to_s
         _queue_change(:dir, dir, rel_path, recursive: true)
       end
 
       def _run_worker(worker)
-        _log(:debug) { "fsevent: running worker: #{worker.inspect}" }
+        Listen.logger.debug { "fsevent: running worker: #{worker.inspect}" }
         worker.run
       rescue
         format_string = 'fsevent: running worker failed: %s:%s called from: %s'
--- a/lib/listen/adapter/darwin.rb
+++ b/lib/listen/adapter/darwin.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'thread'
+require 'listen/thread'
 
 module Listen
   module Adapter
@@ -46,7 +46,10 @@ module Listen
         dirs_to_watch = @callbacks.keys.map(&:to_s)
         Listen.logger.info { "fsevent: watching: #{dirs_to_watch.inspect}" }
         worker.watch(dirs_to_watch, { latency: options.latency }, &method(:_process_changes))
-        @worker_thread = Thread.new { _run_worker(worker) }
+        @worker_thread = Listen::Thread.new("worker_thread") do
+          _log(:debug) { "fsevent: running worker: #{worker.inspect}" }
+          worker.run
+        end
       end
 
       def _process_changes(dirs)
--- a/lib/listen/adapter/darwin.rb
+++ b/lib/listen/adapter/darwin.rb
@@ -80,7 +80,7 @@ module Listen
       end
 
       def _stop
-        @worker_thread.kill.join if (@worker_thread ||= nil)
+        @worker_thread&.kill
         super
       end
     end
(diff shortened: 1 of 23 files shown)


ColinDKelley:
@jonathanhefner Can you please also review?

ColinDKelley:
I just rebased to master.

lib/listen/adapter/darwin.rb
```diff
@@ -44,9 +44,9 @@ def _run
         require 'rb-fsevent'
         worker = FSEvent.new
         dirs_to_watch = @callbacks.keys.map(&:to_s)
-        _log(:info) { "fsevent: watching: #{dirs_to_watch.inspect}" }
+        Listen.logger.info { "fsevent: watching: #{dirs_to_watch.inspect}" }
         worker.watch(dirs_to_watch, { latency: options.latency }, &method(:_process_changes))
-        @worker_thread = Thread.new { _run_worker(worker) }
+        @worker_thread = Listen::Thread.new("worker_thread") { worker.run }
```
ioquatix, line 49: Using the fully qualified name is not strictly necessary although it is possibly better since `Thread.new` might give the wrong impression.

lib/listen/adapter/darwin.rb
```diff
@@ -62,22 +62,22 @@ def _process_changes(dirs)
...
 
       def _process_event(dir, path)
-        _log(:debug) { "fsevent: processing path: #{path.inspect}" }
+        Listen.logger.debug { "fsevent: processing path: #{path.inspect}" }
         # TODO: does this preserve symlinks?
         rel_path = path.relative_path_from(dir).to_s
         _queue_change(:dir, dir, rel_path, recursive: true)
       end
 
       def _run_worker(worker)
```
ioquatix, line 71: Are we still using this method?

lib/listen/logger.rb
```diff
@@ -1,34 +1,35 @@
...
-        case debugging.to_s
-        when /2/
-          ::Logger::DEBUG
-        when /true|yes|1/i
-          ::Logger::INFO
-        else
-          ::Logger::ERROR
-        end
+    def logger
+      @logger ||= default_logger
```
ioquatix, line 13: Maybe not thread safe :p

lib/listen/thread.rb
```diff
@@ -0,0 +1,45 @@
...
+      # Any exceptions raised by the thread will be logged with the thread name and complete backtrace.
+      def new(name)
+        thread_name = "listen-#{name}"
+
+        caller_stack = caller
+        ::Thread.new do
+          begin
+            yield
+          rescue Exception => ex
+            _log_exception(ex, thread_name, caller_stack)
```
ioquatix, line 18: Maybe consider https://rubyapi.org/2.7/o/thread#method-i-report_on_exception

spec/lib/listen/adapter/polling_spec.rb
```diff
@@ -61,7 +61,6 @@
         t = Thread.new { subject.start }
         sleep 0.25
         t.kill
-        t.join
```
ioquatix, line 64: Do you need to call `join` after `kill`?

ioquatix (approved):
Some minor feedback for discussion.

lib/listen/adapter/darwin.rb
ColinDKelley: Yes, that was my thought. `Thread.new` would lead many astray.

lib/listen/adapter/darwin.rb
ColinDKelley: Yes, it looks to me like it's called from the base class:
https://github.com/Invoca/listen/blob/de1e7cb808226ecfaa9bc3dc3a002812e8329fff/lib/listen/adapter/base.rb#L45

lib/listen/logger.rb
ColinDKelley: Yes, I suppose 2 could wind up getting created, but I think we'd be ok, since they're equivalent.

lib/listen/thread.rb
ColinDKelley: Can you clarify your suggestion? Are you suggesting to not log ourselves and instead fall into the Ruby code? Or to check that boolean and do something different here?

spec/lib/listen/adapter/polling_spec.rb
ColinDKelley: No, `kill` is synchronous, so it won't return until the thread is killed. `join` is unnecessary at that point.

lib/listen/thread.rb
ioquatix: I just wonder what is the reason why we use our own thread wrapper, is it sufficient to use `report_on_exception`?

ColinDKelley: Thanks for the clarification. I don't think the `report_on_exception` feature is sufficient. It writes directly to `$stderr`, which is better than nothing, but in general it won't appear in the `logger` with proper formatting. (In kubernetes will wind up swept into the `logger` at least--because they capture `$stderr` and `$stdout`--but in all cases, the log formatting will be missing. No date stamp, no severity. At our company we use JSON logging and of course this won't be formatted as valid JSON.)

I do like that Ruby added the `report_on_exception` feature, but I view it as a safety net that may help you see _something_ if you forgot to rescue and log at the top of the thread. Anecdote: over 10 years ago when we were running in production on Ruby 1.8.7, we had some terrible mystery bugs where "the process just stopping working". We eventually traced these down to exceptions getting raised in some worker threads. Those were intermittent, and getting silently ignored by Ruby, so it took much longer than it should have to figure out! From then on, we _always_ put `rescue`-and-log at the top of our threads and fibers.

Aside: I think the Ruby language should have the idea of a "default logger" for the thread/fiber/process and it should log there and *never* write to $stderr directly. Instead, the default for the "default logger" could do that, with degenerate formatting (don't show datestamp and severity, for example) so it looks about like it looks now when they call `$stderr.puts`. But applications that have their own logging strategy--which every production application does--could set Ruby default logger to point there so Ruby's messages would get plumbed to the logs with proper formatting.

ioquatix: All makes sense.

lib/listen/adapter/darwin.rb
jonathanhefner: I think @ioquatix was referring to `_run_worker`, rather than `_process_event`.  I don't think `_run_worker` is used elsewhere, at least according to `git grep _run_worker`.

ColinDKelley: Oh, thanks for spotting my misunderstanding @jonathanhefner. That is indeed true. This unused `_run_worker` might have been a from a mistaken merge conflict resolution. Deleting it now.

ColinDKelley:
Thanks for the review, @ioquatix and @jonathanhefner. I believe all the comments are addressed, but don't hesitate to @-mention me if I missed anything, or more discussion would be helpful.

Rebasing + merging into master now.

ColinDKelley:
Merged.

PR #503, opened 2020-11-03
Update README for v3.3

ColinDKelley:


ioquatix (approved):
I didn't read the full thing because I'm pretty busy today, but it looks like a big improvement.

ColinDKelley:
Thanks @ioquatix. I had the day off for the election--but had already voted--so it was a good day to work on this gem!

Issue #510, opened 2020-11-07
listen uses Time.now for time intervals; should use Monotonic tick count


ColinDKelley:
## Current State

The `'listen` gem uses the `Time.now` clock for measuring time intervals. For example, from base.rb:

```ruby
      def _wait_until_events_calm_down
        loop do
          now = _timestamp

          # Assure there's at least latency between callbacks to allow
          # for accumulating changes
          diff = _deadline - now
          break if diff <= 0

          # give events a bit of time to accumulate so they can be
          # compressed/optimized
          _sleep(diff)
        end
      end

...
      def _timestamp
        config.timestamp
      end
...
      def timestamp
        Time.now.to_f
      end
```

This is a bug since the `Time.now` timer may be updated at any time, forward or backward. Most commonly by `ntp`.

## Desired State

The `'listen` gem should use the Monotonic tick count instead for time intervals. There are several gems that encapsulate this, including https://github.com/Freaky/monotime, https://github.com/invoca/monotonic_tick_count. Perhaps best for portability and minimal coupling, [concurrent-ruby](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent-ruby/concurrent/utility/monotonic_time.rb) has a simple encapsulation as `monotonic_time`.

## Steps to Reproduce

I didn't try, but you could reproduce the bug by adjusting the system clock while `listen` is running. If you adjusted the clock forward by, say, 1 second, you could trick the `_wait_until_events_calm_down` method into thinking it didn't need to wait at all, thus leading to duplicate events.

ColinDKelley:
@ioquatix Curious to get your opinion here.

ioquatix:
Just use the Ruby interface don't pull in a dependency for it.

ColinDKelley:
Sounds good to avoid dependencies. I wonder if we'll have trouble with some Ruby versions not supporting `Process::CLOCK_MONOTONIC`? `concurrent-ruby` has code that adapts for that not being defined in MRI, and a special case that talks to the JVM in JRuby. I suppose that could just be outdated? If we do hit it, we could copy that `if/elsif/else` logic in here, I suppose.

ioquatix:
You can ignore any version that doesn't support it, they should all be EOL.

ColinDKelley:
Sounds good.
BTW this is a minor bug in the scheme of things, so I'd like to release v3.3.0 before fixing this.

ioquatix:
[...]

PR #512, opened 2020-11-09
Issue #510: use monotonic tick count

ColinDKelley:
Addressed issue #510:
- Added `Listen::MonotonicTime` module with `now` method. including tests.
- Change relative time usage to the above.
- Left absolute time (compared with `mtime`) as it was, using `Time.now`.
- Removed unnecessary `timestamp` and `_timestamp` methods.

### DIFF
--- a/lib/listen/monotonic_time.rb
+++ b/lib/listen/monotonic_time.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Listen
+  module MonotonicTime
+    class << self
+      def now
+        if defined?(Process::CLOCK_MONOTONIC)
+          Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        elsif defined?(Process::CLOCK_MONOTONIC_RAW)
+          Process.clock_gettime(Process::CLOCK_MONOTONIC_RAW)
+        else
+          Time.now.to_f
+        end
+      end
+    end
+  end
+end
--- a/lib/listen/monotonic_time.rb
+++ b/lib/listen/monotonic_time.rb
@@ -3,14 +3,24 @@
 module Listen
   module MonotonicTime
     class << self
-      def now
-        if defined?(Process::CLOCK_MONOTONIC)
+      if defined?(Process::CLOCK_MONOTONIC)
+
+        def now
           Process.clock_gettime(Process::CLOCK_MONOTONIC)
-        elsif defined?(Process::CLOCK_MONOTONIC_RAW)
+        end
+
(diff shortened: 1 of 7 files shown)


lib/listen/monotonic_time.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Listen
+  module MonotonicTime
+    class << self
+      def now
+        if defined?(Process::CLOCK_MONOTONIC)
+          Process.clock_gettime(Process::CLOCK_MONOTONIC)
```
ioquatix, line 8: I can tell you that on every Ruby implementation that I'm aware of, `Process::CLOCK_MONOTONIC` will be defined.

If you insist on this, you should at least put the if statement outside the method definition rather than checking it every time.

ColinDKelley: Since I can never see every Ruby implementation, I feel best going with the documentation.

I moved the `if` outside. That was tricky to get the tests to still pass, but I think I got, by using `load` to keep reloading the file, and especially one final time in a `before(:all)` block. https://github.com/guard/listen/pull/512/commits/0ea75950098ff11e1c217119dd44181a8500c6d8

spec/lib/listen/event/processor_spec.rb
```diff
@@ -132,15 +132,15 @@ def status_for_time(time)
             it 'still does not process events because it is paused' do
               # pretend we were woken up at 0.6 seconds since start
               allow(config).to receive(:sleep).
-                with(anything) { |*_args| state[:time] += 2.0 }
+                with(anything) { |*_args| state[:monotonic_time] += 2.0 }
```
ioquatix, line 135: Is it necessary to rename this field?

ioquatix (changes requested):

spec/lib/listen/event/processor_spec.rb
ColinDKelley: I found it helpful to keep clear that that was not the `Time.now` stub value but rather the `MonotonicTime.now` stub value.
It looked to me like it was just used in this test, so I didn't see a downside to clarifying the name.
Did I get that wrong?

ioquatix: I'm not sure I was just concerned it was part of a public interface.

ColinDKelley: My read is that it's strictly internal to this test. The only reason it's in a hash is so it can be mutated as the test runs, to mimic time advancing. (An `@` variable in the test would work that way too.)

Issue #513, opened 2020-11-13
listener.stop fails with undefined method `alive?' for nil:NilClass if called before started


ColinDKelley:
## Current State

If a listener hasn't been started yet, a call to `stop` will raise an exception:

```
listener = Listen.to(path) do { ... }
listener.stop
NoMethodError:
       undefined method `alive?' for nil:NilClass
     # /usr/local/bundle/gems/listen-3.3.0/lib/listen/event/loop.rb:70:in `stop'
     # /usr/local/bundle/gems/listen-3.3.0/lib/listen/listener.rb:82:in `block in <class:Listener>'
     # /usr/local/bundle/gems/listen-3.3.0/lib/listen/fsm.rb:121:in `instance_eval'
     # /usr/local/bundle/gems/listen-3.3.0/lib/listen/fsm.rb:121:in `call'
     # /usr/local/bundle/gems/listen-3.3.0/lib/listen/fsm.rb:102:in `transition_with_callbacks!'
     # /usr/local/bundle/gems/listen-3.3.0/lib/listen/fsm.rb:69:in `transition'
     # /usr/local/bundle/gems/listen-3.3.0/lib/listen/listener.rb:101:in `stop'
```

## Desired State

The `stop` method should always succeed--even when not started.

## Steps to Reproduce

Call `stop` before starting:
```
listener = Listen.to(path) do { ... }
listener.stop
```

ColinDKelley:
Fixed by https://github.com/guard/listen/pull/514

PR #514, opened 2020-11-13
issue #513: allow stop when not started

ColinDKelley:
Fix issue #513 .

This issue wasn't being caught by the unit tests because they stub the processor entirely. So I added a couple tests to `acceptance_test`.

Note: I changed the indentation to add a new `context`, so this is best reviewed with "hide whitespace changes".

### DIFF
--- a/lib/listen/event/loop.rb
+++ b/lib/listen/event/loop.rb
@@ -64,12 +64,9 @@ module Listen
       end
 
       def stop
-        return if stopped?
         transition! :stopped
 
-        if @wait_thread&.alive?
-          @wait_thread.join
-        end
+        @wait_thread&.join
         @wait_thread = nil
       end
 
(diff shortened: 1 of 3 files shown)


lib/listen/event/loop.rb
```diff
@@ -67,7 +67,7 @@ def stop
         return if stopped?
         transition! :stopped
 
-        if @wait_thread.alive?
+        if @wait_thread&.alive?
```
ColinDKelley, line 70: Feeback from @ioquatix 

> I think you can just do @wait_thread.join
> the if statement, is a race condition

Building on that, the `return if stopped?` is too. Getting rid of both, in favor of simply `@wait_thread&.join`.

ioquatix (approved):

PR #515, opened 2020-11-22
Fix description typo

ccouzens:
It was showing up in https://rubygems.org/gems/listen


ColinDKelley:
@ccouzens Thank you for fixing this! Will release with 3.3.2.

PR #518, opened 2020-11-27
feat: Add rubocop lint github action

### LINKED ISSUE #516: Replace Hound with GitHub Action
Hound doesn't seems to be maintained as the maximum supported `rubocop` version is `0.91`.

As an alternative we can use GitHub Actions which have emerged as a built-in solution for this kind of tasks.
There are many github actions that support annotation based linting, eg: [lint-action](https://github.com/wearerequired/lint-action) which can be replaced with hound.

However `lint-action` suffers from a restriction in which a PR from a fork failed to add annotations, for which `pull_request_target` event was added, but it has also failed (for [me](https://github.com/KapilSachdev/shoulda-matchers/runs/1407720196)).

1. We can look for similar github actions which are working fine using `pull_request_target` event. OR
2. Can add a non-annotation based linting action for now, [see here](https://github.com/rails/rails/runs/1445160196)

KapilSachdev:
- Remove hound

This PR adds non-annotation based rubocop lint action.
Closes #516 

Currently in draft. Looking for a stable annotation based linting action.

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -1,0 +1,20 @@
+name: RuboCop
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v2
+
+    - name: Setup Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 2.7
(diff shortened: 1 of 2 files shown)


ColinDKelley:
@KapilSachdev Since this PR is over a year old, I'd like to close it now. You can always open a new PR if you have time to make one that completes the work.

PR #519, opened 2020-11-29
Add project metadata to the gemspec

orien:
Add `bug_tracker_uri`, `changelog_uri`, `documentation_uri`, `homepage_uri`, `source_code_uri`, and `wiki_uri` to the gemspec metadata.

These [project metadata](https://guides.rubygems.org/specification-reference/#metadata) will facilitate easy access to project information. The URI will be available on the [Rubygems project page](https://rubygems.org/gems/listen), via the rubygems API, and the `gem` and `bundle` command-line tools with the next release.

### DIFF
--- a/listen.gemspec
+++ b/listen.gemspec
@@ -16,7 +16,13 @@ Gem::Specification.new do |gem|
   gem.description = 'The Listen gem listens to file modifications and '\
     'notifies you about the changes. Works everywhere!'
   gem.metadata = {
-    'allowed_push_host' => 'https://rubygems.org'
+    'allowed_push_host' => 'https://rubygems.org',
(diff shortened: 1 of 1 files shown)


ColinDKelley:
@orien Thank you for the PR. Just curious: is there a reference list for those specific metadata keys?

orien:
Yes there is. https://guides.rubygems.org/specification-reference/#metadata

ColinDKelley:
Oh, ok! I had already clicked through to that list but it seemed to be just some random examples; I didn't realize it was a reference list for the full set of supported keys. But I guess it is.

I'll merge now. This should ship with [v3.4](https://github.com/guard/listen/milestone/10).

PR #521, opened 2020-12-01
put missing wiki content into README

ColinDKelley:
The wiki hadn't been maintained in many years, so I deleted it. There was still some content referred to by the README, however. This copies that content inline.


seandunn:
That's perfect, thanks @ColinDKelley.

Issue #473, opened 2020-03-15
Improve wiki on "Increasing the amount of inotify watchers"


AxelTheGerman:
Hi,
I used Rails for years but now switched to Linux and for the first time ran into the issue of
```
FATAL: Listen error: unable to monitor directories for changes.
Visit https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers for info on how to fix this.
```

Thanks for having great documentation, however I was wondering if we should improve it.

It says for ArchLinux to write the new config to `/etc/sysctl.d/40-max-user-watches.conf` but on my machine I already had `/etc/sysctl.d/50-max_user_watches.conf` present which continued to overwrite the new limit I was trying to set.

Maybe we should add a note to check the `/etc/sysctl.d/` directory for other config files that would overwrite the setting again? Maybe even add this `grep` command to check the users existing configuration:
```
$ grep -H -s "fs.inotify.max_user_watches" /etc/sysctl.d/*
/etc/sysctl.d/50-max_user_watches.conf:fs.inotify.max_user_watches=524288
```

Thanks!

ColinDKelley:
@AxelTheGerman 
Those sound like good improvements to make to the docs. Can you make a PR?

seandunn:
Just to note that the error message given in the terminal currently directs anyone hitting this issue to the main README which doesn't include any information on dealing with it.

ColinDKelley:
Oops, before deleting the wiki, I meant to copy the references from wiki into the README.

@seandunn How does this look?

https://github.com/guard/listen/pull/521

seandunn:
That's perfect, thanks @ColinDKelley.

PR #522, opened 2020-12-01
issue #473: update README for setting fs.inotify.max_user_watches

ColinDKelley:
- Issue #473 : Update README for setting `fs.inotify.max_user_watches`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,13 +295,19 @@ $ bundle exec sass --watch # ... or whatever app is using Listen.
 
 If you are running Debian, RedHat, or another similar Linux distribution, run the following in a terminal:
 ```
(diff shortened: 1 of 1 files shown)


ColinDKelley:
@AxelTheGerman How does this look? I tried to simplify the steps by using `sudo echo >`/`sudo echo >>` rather than `sudo tee`/`sudo tee -a`.

ColinDKelley:
I'm going to merge now. Happy to get any feedback later.

JasonBarnabe:
The new instructions don't work. `sudo` does not apply to the redirection.

```
$ sudo echo fs.inotify.max_user_watches=524288 >> /etc/sysctl.conf
bash: /etc/sysctl.conf: Permission denied
```

Issue #451, opened 2018-08-09
wdm install advice on Windows could be improved


belluzj:
I'm working on a Rails app and developing on Windows. I got advice to add `wdm` to my Gemfile and did as instructed, but this resulted later in a problem when deploying the same Rails app on a Linux machine, where bundler complained:

```
You are trying to install in deployment mode after changing
your Gemfile. Run `bundle install` elsewhere and add the
updated Gemfile.lock to version control.

If this is a development machine, remove the /usr/src/app/Gemfile freeze 
by running `bundle config --delete frozen`.

You have deleted from the Gemfile:
* wdm (>= 0.1.0)
```

I fixed the issue by changing my Gemfile:

```diff
-gem 'wdm', '>= 0.1.0' if Gem.win_platform?
+gem 'wdm', '>= 0.1.0', platforms: [:x64_mingw]
```

From what I understand, that way bundler knows that the gem is only needed on windows and can be smart about it. Maybe you could update the advice message that you display? However the platform symbol (`:x64_mingw`) may be different for each Ruby-on-Windows install.

FrankFang:
Thanks!

PR #523, opened 2020-12-01
issue #451: change windows install instructions to suggest platforms: instead of Gem.win_platform?

ColinDKelley:
- Issue #451 : change Windows install instructions to suggest `platforms:` instead of `Gem.win_platform?`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ If you are on Windows, it's recommended to use the [`wdm`](https://github.com/Ma
 Please add the following to your Gemfile:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


ColinDKelley:
@belluzj : how does this look?

belluzj:
Thanks for looking into this! Sorry but I won't be able to reproduce my original issue now; however the patch looks fine from my recollection of the problem.

ColinDKelley:
Thanks, I'll merge it now. We can always improve it later.

PR #524, opened 2020-12-11
Don't return incorrect files when there's a file whose name matches a dir

ghiculescu:
Fixes a bug in https://github.com/guard/listen/pull/460

We have a file named `cypress.json`, and a dir named `cypress`, in a repo we are watching with `listen`.

Since upgrading to 3.3, `listen` will get confused and report that the file `cypress/.json` was removed. No such file exists.

I'm pretty sure this PR fixes that.

### DIFF
--- a/lib/listen/record.rb
+++ b/lib/listen/record.rb
@@ -59,13 +59,13 @@ module Listen
 
     def _sub_tree(rel_path)
       @tree.each_with_object({}) do |(path, meta), result|
-        next unless path.start_with?(rel_path)
-
-        if path == rel_path
-          result.merge!(meta)
-        else
-          sub_path         = path.sub(%r{\A#{rel_path}/?}, '')
-          result[sub_path] = meta
+        parts = path.split(::File::SEPARATOR)
+        if parts.shift == rel_path
+          if parts.empty?
+            result.merge!(meta)
+          else
+            result[parts.join(::File::SEPARATOR)] = meta
+          end
         end
       end
     end
--- a/spec/lib/listen/record_spec.rb
+++ b/spec/lib/listen/record_spec.rb
@@ -208,6 +208,29 @@ RSpec.describe Listen::Record do
       end
     end
 
+    context 'when there is a file with the same name as a dir' do
+      subject { record.dir_entries('cypress') }
+
+      before do
+        record.update_file('cypress.json', mtime: 1.1)
+        record.update_file('cypress/README.md', mtime: 1.2)
+        record.update_file('a/b/cypress/d', mtime: 1.3)
+        record.update_file('a/b/c/cypress', mtime: 1.3)
+      end
+      it { should eq('README.md' => { mtime: 1.2 }) }
+    end
+
+    context 'when there is a file with a similar name to a dir' do
+      subject { record.dir_entries('app') }
+
+      before do
+        record.update_file('appspec.yml', mtime: 1.1)
+        record.update_file('app/README.md', mtime: 1.2)
+        record.update_file('spec/app/foo', mtime: 1.3)
+      end
+      it { should eq('README.md' => { mtime: 1.2 }) }
+    end
+
     context 'in subdir /path' do
       subject { record.dir_entries('path') }
 


lib/listen/record.rb
```diff
@@ -64,7 +64,10 @@ def _sub_tree(rel_path)
         if path == rel_path
           result.merge!(meta)
         else
-          sub_path         = path.sub(%r{\A#{rel_path}/?}, '')
+          parts = path.split("/")
```
ColinDKelley, line 67: I think we should use `File::SEPARATOR`, so it works in Windows too.

lib/listen/record.rb
```diff
@@ -64,7 +64,10 @@ def _sub_tree(rel_path)
...
+          next unless parts.include?(rel_path)
```
ColinDKelley, line 68: Hmm, can `rel_path` ever have separators in it? If not, then this would always have to be true, wouldn't it, since we established a few lines up that it _starts with_ `rel_path`?

lib/listen/record.rb
```diff
@@ -64,7 +64,10 @@ def _sub_tree(rel_path)
...
+
+          sub_path = (parts - [rel_path]).join("/")
```
ColinDKelley, line 70: This will remove `rel_path` from anywhere in the path. Wouldn't that trip up with something like
```
a/b/c/cypress
```
(a file that happened to have that same name)
or
```
a/b/cypress/d
```
(a nested folder that happened to have that same name)

ColinDKelley: And the `join` could use `File::SEPARATOR` too.

lib/listen/record.rb
```diff
@@ -64,7 +64,10 @@ def _sub_tree(rel_path)
         if path == rel_path
           result.merge!(meta)
         else
-          sub_path         = path.sub(%r{\A#{rel_path}/?}, '')
```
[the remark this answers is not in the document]
ColinDKelley: Also it was a bug that the `rel_path` wasn't escaped before interpolating it into a regexp. All sorts of weird things could happen if `rel_path` had characters in it that are special in regexps, like ^ $ [ \ ? | . * etc.

ColinDKelley (commented):
@ghiculescu 
Thanks for filing this fix. I wish I knew the context of this code. Are we only concerned with file patterns that _start_ with `rel_path`? The old code was, but the new code can match (and remove) `rel_path` from anywhere in the path.

ghiculescu:
Hey @ColinDKelley, thanks for the review!

Context is that `listen` is used by https://github.com/daveallie/entangler, we use it to 2-way sync files between local & remote workspaces. So the examples in my test are real examples from my codebase.

lib/listen/record.rb
ghiculescu: Without this line, a bunch of (my new) tests fail.

lib/listen/record.rb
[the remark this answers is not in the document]
ghiculescu: Yeah, I generally find regexes super hard to grok, so I'm a fan of removing it in favour of more explicit code. That's a matter of personal preference though - I can go back to a regex-based solution if you prefer.

lib/listen/record.rb
ghiculescu: I added another test for this case, but it seems to work fine.

ghiculescu: My understanding is this should only return files at `#{rel_path}/**/**`.

lib/listen/record.rb
ghiculescu: > If not, then this would always have to be true, wouldn't it, since we established a few lines up that it starts with `rel_path`?

No, because for example, `"appspec.yml"` starts with `"app"`, but `["appspec.yml"]` doesn't *include* `"app"`.

lib/listen/record.rb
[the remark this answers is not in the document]
ColinDKelley: I agree with you! It is best to avoid regexps in code. My experience with them is that the likelihood of being 100% correct drops quadratically with every punctuation character used. :)

lib/listen/record.rb
ColinDKelley: Ah, you are correct. I was thinking `Array#-` worked as a set, but the docs are clear that it must match from the front, and that you need to a `Set` if you want `#-` to work like a set.

ColinDKelley: Wait, I take that back.
```
irb> ['a', 'b', 'c'] - ['b']
=> ['a', 'c']

You only want to remove from the front, right?

File: lib/listen/record.rb (hunk 2 above)
Comment (ColinDKelley) [reply lib/listen/record.rb:68]: Oh jeez, I missed that too. Can we get rid of all the confusing special cases and simplify down to
```ruby
    def _sub_tree(rel_path)
      @tree.each_with_object({}) do |(path, meta), result|
        parts = path.split(::File::SEPARATOR)
        if parts.shift == rel_path
          if parts.empty?
            result.merge!(meta)
          else
            result[parts.join(::File::SEPARATOR)] = meta
          end
        end
      end
```

lib/listen/record.rb
ghiculescu: Oh I see what you mean. Yeah i suspect you’re right. Let me try your simpler looking code tomorrow.

lib/listen/record.rb
ghiculescu: Okay, I've moved to this implementation.

PR #525, opened 2020-12-20
issue #473: use sudo sh -c so redirection works

ColinDKelley:
@JasonBarnabe Thanks for pointing out that `sudo` wouldn't would with `>` or `>>` after. `sudo sh -c` seems like the best work-around.

How does this look?


JasonBarnabe:
Looks good to me.

ColinDKelley:
Thanks, @JasonBarnabe.

PR #526, opened 2020-12-21
Don't return incorrect files when there's a file whose name matches a dir

ghiculescu:
Simpler attempt at https://github.com/guard/listen/pull/524. Reverts https://github.com/guard/listen/pull/460.

Even with the changes in https://github.com/guard/listen/pull/460 reversed, the acceptance tests still pass. Only one unit test fails, and I'm not sure it's asserting the right behavior. With this approach, the new tests in https://github.com/guard/listen/pull/524 also still pass.

cc @bryanlira @ColinDKelley

### DIFF
--- a/spec/lib/listen/record_spec.rb
+++ b/spec/lib/listen/record_spec.rb
@@ -208,6 +208,29 @@ RSpec.describe Listen::Record do
       end
     end
 
+    context 'when there is a file with the same name as a dir' do
+      subject { record.dir_entries('cypress') }
+
+      before do
+        record.update_file('cypress.json', mtime: 1.1)
+        record.update_file('cypress/README.md', mtime: 1.2)
+        record.update_file('a/b/cypress/d', mtime: 1.3)
+        record.update_file('a/b/c/cypress', mtime: 1.3)
+      end
+      it { should eq('README.md' => { mtime: 1.2 }) }
+    end
+
+    context 'when there is a file with a similar name to a dir' do
+      subject { record.dir_entries('app') }
+
+      before do
+        record.update_file('appspec.yml', mtime: 1.1)
+        record.update_file('app/README.md', mtime: 1.2)
+        record.update_file('spec/app/foo', mtime: 1.3)
+      end
+      it { should eq('README.md' => { mtime: 1.2 }) }
+    end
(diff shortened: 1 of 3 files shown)


spec/lib/listen/record_spec.rb
```diff
@@ -220,9 +243,17 @@ def record_tree(record)
         it { should eq('file.rb' => { mtime: 1.1 }) }
       end
 
-      context 'with path/subdir' do
+      context 'with empty path/subdir' do
         before { record.add_dir('path/subdir') }
-        it { should eq('subdir' => {}) }
```
ghiculescu, line 225: @bryanlira you added this test in https://github.com/guard/listen/pull/460. Do you remember why you expected it to return this, vs an empty hash? The [acceptance test](https://github.com/guard/listen/pull/460/files#diff-cb0c4f7e5adeaa9e4106f4a3f5a73261de4e95b52f0367868a3dd40a2e74b9efR198) you added in that PR is still passing with this branch, so I'm not sure if this unit test was critical for something else that I'm not seeing it.

ColinDKelley:
Nice, @ghiculescu, I certainly like the simplicity from reverting the code. Let's see how @bryanlira feels about that one spec change.

ColinDKelley:
BTW this appears a few lines above the first changes:
```ruby
subtree = if [nil, '', '.'].include? rel_path.to_s
```
There's no way `.to_s` ever returns `nil`, so this could be simplified to:
```ruby
subtree = if ['', '.'].include?(rel_path.to_s)
```

ColinDKelley:
Let's wait another day or so to hear back from @bryanlira. If we don't hear back, I will assume the spec change is good and release this in v3.3.4.

ghiculescu:
Thanks! In the meantime, do you want me to squash my commits or are you happy with the PR as it stands? (I know some projects don't like the "merge and squash" button in github (though I'm not sure why) so I figure I may as well get everything else ready for merge.)

ColinDKelley:
I think this branch is good as is. I know @ioquatix prefers the Rebase-and-Merge options so there are no merge commits... Github is saying this branch is ready to go with that strategy.

PR #527, opened 2020-12-31
added correct link to help debug inotify workers error on linux

williamkennedy:
Pretty straightforward change. 

On my CI build, the INOTIFY_LIMIT_MESSAGE was being displayed but the link was taking me to the wrong place. 

It had me scratching my head for a few seconds until I read the readme. 

Hope it helps.

### DIFF
--- a/lib/listen/adapter/linux.rb
+++ b/lib/listen/adapter/linux.rb
@@ -21,8 +21,9 @@ module Listen
 
       private
 
+      
       WIKI_URL = 'https://github.com/guard/listen'\
-        '/wiki/Increasing-the-amount-of-inotify-watchers'
+        '/blob/master/README.md#increasing-the-amount-of-inotify-watchers'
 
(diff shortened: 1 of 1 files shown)


ColinDKelley:
Oops, I accidentally broke that when I phased out the Wiki in favor of the README. Thanks for the fix!

I'll release this as part of v3.3.4 in the next day or two.

williamkennedy:
> Oops, I accidentally broke that when I phased out the Wiki in favor of the README. Thanks for the fix!
> 
> I'll release this as part of v3.3.4 in the next day or two.

Awesome. I saw that the build was failing on macos and I was not sure why. Was going to have a look at it later but looks like you beat me to it.

ColinDKelley:
@williamkennedy Glad you didn't spend time on the MacOS failure. It's just a slightly flakey test that seems to fail maybe 10% of the time. I've never seen it fail twice in a row.

I released both v3.3.4 and v3.4.0 just now. I separated them because v3.4.0 has a bit more risk to it (dependence on Linux-variant-specific monotonic clocks).

PR #529, opened 2021-01-13
Correctly pass timeout as kwargs to wait_for_state

the-spectator:
### Issue
Related issue: https://github.com/socketry/falcon/issues/138

### Summary of Change
This PR correctly pass `timeout` as kwargs to method call `wait_for_state`.


ColinDKelley:
Sorry about this bug. Looks like that timeout is getting interpreted as another state to wait for. (Hence no test frailures.)

I'm working on adding a couple tests now...

PR #531, opened 2021-01-14
Add tests for PR #529

ColinDKelley:
Builds on PR #529 :
- Adds missing tests to `fsm_spec` (ensure state names are symbols in `wait_for_state`) and `loop_spec` (raise exception if not started in 5 seconds).
- Moves `Error` down to a common `Listen::Error` base class.
- Removes stray references to the wiki since we have switched over to the README.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ Contribute to Listen
 File an issue
 -------------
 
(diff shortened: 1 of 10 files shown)


ColinDKelley:
@the-spectator Can you please review? I added the missing tests that would have caught the #529 bug.

the-spectator (approved):

PR #536, opened 2021-03-29
Bump required ruby version to v2.4.0

cgunther:
a9137acc introduced a use of `Hash#transform_values` in `Listen::Record#dir_entries`, however that was only added to Ruby in [v2.4](https://github.com/ruby/ruby/blob/v2_4_0/NEWS#core-classes-updates-outstanding-ones-only-).

This can lead to runtime errors when run on an older version of Ruby, such as 2.2.7, which the gemspec previously implied was supported.

It looks like CI only runs on Ruby 2.5+, so an argument could also be made the gemspec should follow suit:
https://github.com/guard/listen/blob/c00a57bffa057c520c5b7285420c21e3aff11454/.github/workflows/development.yml#L17-L19

Bumping to v2.4 was the simplest to resolve the immediate problem I was facing running an old project on Ruby v2.3.4 while I work to upgrade it.

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -6,21 +6,22 @@ jobs:
   test:
     runs-on: ${{matrix.os}}-latest
     continue-on-error: ${{matrix.experimental}}
-    
+
     strategy:
       matrix:
         os:
           - ubuntu
           - macos
-        
+
         ruby:
+          - 2.4
           - 2.5
           - 2.6
           - 2.7
-        
+
         experimental: [false]
         env: [""]
-        
+
         include:
           - os: macos
             ruby: truffleruby-head
@@ -34,14 +35,14 @@ jobs:
           - os: ubuntu
             ruby: head
             experimental: true
-    
+
     steps:
     - uses: actions/checkout@v2
     - uses: ruby/setup-ruby@v1
       with:
         ruby-version: ${{matrix.ruby}}
         bundler-cache: true
-    
+
     - name: Run tests
       timeout-minutes: 6
       run: |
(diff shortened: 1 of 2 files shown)


ColinDKelley:
@cgunther Thanks for reporting this bug. I fear it might cause problems in other projects if we narrowed the supported Ruby version.

I believe we can get equivalent behavior to
```
      subtree.transform_values do |values|
        # only get data for file entries
        values.has_key?(:mtime) ? values : {}
      end
```
with
```
      subtree.each_with_object({}) do |(key, values), result|
        # only get data for file entries
        result[key] = values.has_key?(:mtime) ? values : {}
      end
```
Look good?

I'd like to add Ruby 2.2 to the CI matrix as well, if we can.

ColinDKelley:
I'm trying this branch to see how hard it is to get 2.2 tests passing: https://github.com/guard/listen/pull/537

Looks like `rubocop` 0.91 requires ruby 2.4...

cgunther:
Updated!

I'd agree there's little value in supporting Ruby < v2.4. In my little analysis, listen v3.3.2 was the first to use this new Ruby 2.4 feature and that was released in Nov 2020, so if in 4 months no one else complained about an incompatibility with Ruby < 2.4, it may not be much of an issue. From a Ruby perspective, v2.4 reached end-of-life in April 2020, so it's debatable how much effort should go into supporting it, but this PR is at least a step in the right direction.

cgunther:
Thanks for the quick merge!

ColinDKelley:
Released in `listen` v3.5.1.

Thanks again for reporting and fixing this!

PR #538, opened 2021-04-23
BSD: Don't send ignored paths to kqueue

jcs:
I'm using this via Jekyll's `--watch` option and on OpenBSD I either have to get a CPU-eating poll of its giant directory structure (because of `.git` and `vendor/bundle`) or I use `rb-kqueue` and the process quickly dies after running out of file handles.

Since the paths being ignored by the Silencer aren't used anyway, can it just not pass them to kqueue and avoid watching them?

### DIFF
--- a/lib/listen/adapter/bsd.rb
+++ b/lib/listen/adapter/bsd.rb
@@ -45,7 +45,11 @@ module Listen
         @callback = callback
         # use Record to make a snapshot of dir, so we
         # can detect new files
-        _find(directory.to_s) { |path| _watch_file(path, @worker) }
+        _find(directory.to_s) do |path|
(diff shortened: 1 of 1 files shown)


ColinDKelley:
@jcs Thank you for the PR.

Can you please add tests? At least for these cases:
- silenced by :dir 
- silenced by :file
- not silenced

ColinDKelley:
Hi @jcs Just checking in here to see if you can write those 3 tests?

ColinDKelley:
Hi @jcs Just checking in again to see if you can write those 3 tests? LMK if you need help with them.

ColinDKelley:
@jcs Can you check out v3.7.0 that was just released and see if the optimization from #542 winds up accomplishing the same `kqueue` optimization?

jcs:
It appears so, thanks!

ColinDKelley:
Excellent! Thanks @jcs for looking into this.

Issue #452, opened 2018-08-12
Listen's use of Digest::MD5 breaks support on FIPS enabled machines


ruckc:
Please review usage of Digest::MD5 and migrate to a different hash algorithm that is not disabled in when FIPS is enabled.

```
$ rails server
=> Booting Puma
=> Rails 5.2.1 application starting in development
=> Run `rails server -h` for more startup options
md5_dgst.c(78): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
Aborted
```

ioquatix:
What is the best solution to this problem?

ruckc:
Replacing Digest::MD5 with a modern hashing algorithm such as Digest::SHA2 is the easy answer.  Long term though, its ideal to support configurable algorithms for future proofing.

souravb65:
Is there a plan in the near future to use Digest::SHA2 and make it FIPS enabled?

ioquatix:
Sure, do you have time to submit a PR?

souravb65:
Sure. I can give it a try. Looks like I don't have permission to create my feature branch. Can you please grant me permission?

ioquatix:
Please create a fork and submit a PR. If you need more help, please let me know.

ColinDKelley:
@souravb65 Here's a branch: https://github.com/guard/listen/pull/541 Tests are green!

ColinDKelley:
Considering this fixed.

souravb65:
Awesome @ColinDKelley making these changes. I did the changes but due to my machine issue was not able to test to push those.

ColinDKelley:
No problem, @souravb65. Thanks for the nudge to get this fixed and closed.

ioquatix:
@ColinDKelley Thanks for your effort maintaining this gem.

PR #541, opened 2021-07-20
issue #452: use SHA256 instead of MD5 for FIPS compatibility

ColinDKelley:
Addresses issue #452 by using SHA256 instead of MD5.

### DIFF
--- a/lib/listen/file.rb
+++ b/lib/listen/file.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'digest/md5'
+require 'digest'
 
 module Listen
   class File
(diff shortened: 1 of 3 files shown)


ioquatix (approved):
Oh no.. `MD5`.

ColinDKelley:
Thanks @ioquatix.

What do you think about version bumping for this? It's not a major version bump, since no interfaces changed. But the dependencies did change slightly, so probably a minor version bump?

ioquatix:
I think it's fine. No one should be depending on the checksum implementation.

ColinDKelley:
> No one should be depending on the checksum implementation.

I agree they shouldn't. But if you told me that Truffle- or J- or MRI Ruby version X has a problem where the `SHA256` interface is broken in some way (perhaps because of `libssl`?) I'd believe you in an instant. I've started leaning towards minor version bumps for dependency changes for this reason.

But here's an even better reason: this release actually adds a feature: FIPS compatibility.

PR #542, opened 2021-07-22
fix: Avoid scanning and building entries for silenced directories

ElMassimo:
### Description 📖 

This pull request patches the `Record#build` method to avoid recursively scanning and creating entries for directories that are being ignored in the `Listener` configuration.

### Background 📜 

Building entries that will be ignored by the listener degrades the initialization performance, and can greatly increase memory consumption.

This is especially noticeable on projects with large `.git` or `node_modules` directories.

### The Fix 🔨 

Use the `Silencer` configuration to avoid scanning ignored directories.

### Benchmarks 📈 

The performance difference is significant on startup time.

Using the `listen` gem in a [small Jekyll site](https://github.com/ElMassimo/jekyll-vite/tree/main/docs):

#### Before (silenced dirs are scanned)
```
Record.build(): 0.70397 seconds
```

#### After (silenced dirs are ignored)
```
Record.build(): 0.02439 seconds
```

### Additional Benefits ✨ 

Directories that are ignored would no longer output symlink warnings since they are no longer being scanned.

### DIFF
--- a/lib/listen/adapter/base.rb
+++ b/lib/listen/adapter/base.rb
@@ -51,7 +51,7 @@ module Listen
         # TODO: separate config per directory (some day maybe)
         change_config = Change::Config.new(config.queue, config.silencer)
         config.directories.each do |dir|
-          record = Record.new(dir)
+          record = Record.new(dir, config.silencer)
           snapshot = Change.new(change_config, record)
           @snapshots[dir] = snapshot
         end
--- a/lib/listen/record.rb
+++ b/lib/listen/record.rb
@@ -11,9 +11,10 @@ module Listen
 
     attr_reader :root
 
-    def initialize(directory)
+    def initialize(directory, silencer)
       @tree = _auto_hash
       @root = directory.to_s
+      @silencer = silencer
     end
 
     def add_dir(rel_path)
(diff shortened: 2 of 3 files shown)


ColinDKelley (commented):
@ElMassimo Thanks for the PR. Sorry about the delayed response.

It looks like you filled in a test with an empty silencer. But is there also a test where a silencer is configured, to prove that the short-circuiting `return` happens?

ElMassimo:
@ColinDKelley Hi! Not sure if I recall correctly, but the silencer in the test is not empty.

The silencer always uses some [defaults](https://github.com/guard/listen/blob/master/lib/listen/silencer.rb#L7), without this patch `.git` would show up in the [record tree](https://github.com/guard/listen/pull/542/files#diff-4edd115ec495a5757fc10b52b84fd3ed6dc354ec5aabc27bf23463e495fa8db3R323-R328).

ColinDKelley:
@ElMassimo Aha! Glad it's being tested.
I think it would be an improvement if the test here didn't depend on the defaults of another class. That's an easy change. I'll make it in a PR after I merge this, and tag you.

ColinDKelley:

PR #544, opened 2021-08-14
Plumb Silencer.new and use it #542 spec

ColinDKelley:
- Use explicit ` ignore!: [/\A\.ignored/]` silencer option in `record_spec` so we're not depending on `Silencer`'s defaults.
- Update `Silencer` to allow its config to be passed into `initialize` rather than mutated by `configure`. Update some comments there about deprecating mutation.
- Update a few stubs that no longer matched.

### DIFF
--- a/lib/listen/silencer.rb
+++ b/lib/listen/silencer.rb
@@ -57,23 +57,21 @@ module Listen
       | ~
     )$}x.freeze
 
+    # TODO: deprecate these mutators; use attr_reader instead
     attr_accessor :only_patterns, :ignore_patterns
 
-    def initialize
-      configure({})
+    def initialize(options = {})
+      configure(options)
     end
 
+    # TODO: deprecate this mutator
     def configure(options)
       @only_patterns = options[:only] ? Array(options[:only]) : nil
       @ignore_patterns = _init_ignores(options[:ignore], options[:ignore!])
     end
 
-    # Note: relative_path is temporarily expected to be a relative Pathname to
-    # make refactoring easier (ideally, it would take a string)
-
-    # TODO: switch type and path places - and verify
     def silenced?(relative_path, type)
-      path = relative_path.to_s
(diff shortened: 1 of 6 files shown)


ColinDKelley:
@ElMassimo Here's the follow-up PR from #542. Look good?

ElMassimo:
Excellent, thanks Collin!

ColinDKelley:
Merging this now since few tests are broken without it.

ColinDKelley:

lib/listen/silencer.rb
```diff
@@ -57,23 +57,21 @@ class Silencer
       | ~
     )$}x.freeze
 
+    # TODO: deprecate these mutators; use attr_reader instead
     attr_accessor :only_patterns, :ignore_patterns
 
-    def initialize
-      configure({})
+    def initialize(options = {})
```
ioquatix, line 63: I'm not sure the backwards compatibility constraints but you should prefer `**options`.

ioquatix (commented):
LGTM.

lib/listen/silencer.rb
ColinDKelley: All previous cases initialized this object with no arguments, so I don't think there's a concern there.

Can you help me understand what cases will work better with `**options` vs. `options = {}`?

ColinDKelley: I know that `**options` is effective at collecting the options hash as the last argument when there are many arguments. But with a single argument like this, I'm not seeing a difference between `**options` and `options = {}`?

ColinDKelley: Maybe `**options` is preferred because it always works regardless of other arguments before it?
I switched over to it here: https://github.com/guard/listen/commit/45921334d45e7c09bfca4c29c6e3abd8ba79e632

Issue #509, opened 2020-11-06
linux driver can abort the enclosing process


ColinDKelley:
## Current State

If `:INotify::Notifier.new` (or `#watch`) were to raise `Errno::ENOSPC`, the linux driver would `abort` the containing process:

```ruby
      def _configure(directory, &callback)
        require 'rb-inotify'
        @worker ||= ::INotify::Notifier.new
        @worker.watch(directory.to_s, *options.events, &callback)
      rescue Errno::ENOSPC
        abort(INOTIFY_LIMIT_MESSAGE)
      end
```

Calling `abort` from a gem is never appropriate.

## Desired State

`Errno::ENOSPC` should be wrapped by raising a named exception (with `ENOSPC` as the cause). That will raise out of the public interface `start`. The calling code in the containing process can then take the appropriate action, by rescuing that specific exception or a catch-all base class.

## Steps to Reproduce

I didn't try, but you'd need to run out the count of `inotify` file descriptors. Then call `listener.start`. That would cause the containing process to abruptly `abort`.

ColinDKelley:
@ioquatix Do you agree that this should be fixed? I've never seen it happen in production...but we'd be appalled if it did. It could cause a total outage of a company's service! Yikes. This feels like a "ticking time bomb".

ioquatix:
Why was it added in the first place?

ColinDKelley:
That line hasn't been touched in 7 years: https://github.com/guard/listen/commit/f9e82202d80c84f518918cc199cafc804286a6f0

ColinDKelley:
But it appears to have first been added 8.5 years ago: https://github.com/guard/listen/commit/5e3053d22b7f2ced2230e1192a3426c223056704

PR #545, opened 2021-08-15
issue #509: raise Listen::Error::INotifyMaxWatchesExceeded rather than abort

ColinDKelley:
Fix for issue #509:
- When iNotify max watchers exceeded, raise `INotifyMaxWatchesExceeded` rather than abort.

### DIFF
--- a/lib/listen/adapter/linux.rb
+++ b/lib/listen/adapter/linux.rb
@@ -24,17 +24,15 @@ module Listen
       README_URL = 'https://github.com/guard/listen'\
         '/blob/master/README.md#increasing-the-amount-of-inotify-watchers'
 
-      INOTIFY_LIMIT_MESSAGE = <<-EOS
-        FATAL: Listen error: unable to monitor directories for changes.
-        Visit #{README_URL} for info on how to fix this.
-      EOS
-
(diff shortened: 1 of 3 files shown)


ColinDKelley:

ioquatix (approved):
It looks okay to me. However, I'm not sure if the exception message needs to be so elaborate?

ColinDKelley:
@ioquatix I carried the exception across from the old message. I like that it points to the documentation. How would you simplify it? Were you thinking of putting it all on one line maybe?

ioquatix:
I think the exception message shouldn't contain documentation. So, I'd just say "Max watches exceeded!".

Personally, I don't mind either approach, but if you start adding URLs to exception messages, surely that should apply to all exceptions? So, it's hard to know where to draw the line. Perhaps having documentation on the exception class which explains what to do to mitigate it.

ColinDKelley:
I don't see any downside to the exception including links to documentation on how to resolve the problem! The reason we don't do it all the time is that it's a lot of work. ;)  In this case, someone before me already did it, and I confirmed that the link still works (after I refactored it into the README) so I think we're good. I will trim it down to the more succinct version, though.

PR #546, opened 2021-08-19
Ignore emacs backup/swap files by default.

### LINKED ISSUE #543: Ignore emacs editor temp file.
Emacs will generate some temp files before save, e.g. `.#app.rb`, `#app.rb#`.

So, i thought we should add those rarely used filename to ignore list. like this:

```rb
listener.ignore /\A\.?#/
```

zw963:
For fix #543

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ listener.stop    # stop both listening to changes and processing them
 
 ### Ignore / ignore!
 
-`Listen` ignores some directories and extensions by default (See DEFAULT_IGNORED_DIRECTORIES and DEFAULT_IGNORED_EXTENSIONS in Listen::Silencer).
+`Listen` ignores some directories and extensions by default (See DEFAULT_IGNORED_FILES and DEFAULT_IGNORED_EXTENSIONS in Listen::Silencer).
 You can add ignoring patterns with the `ignore` option/method or overwrite default with `ignore!` option/method.
 
 ``` ruby
@@ -157,7 +157,7 @@ All the following options can be set through the `Listen.to` after the directory
 
 ```ruby
 ignore: [%r{/foo/bar}, /\.pid$/, /\.coffee$/]   # Ignore a list of paths
-                                                # default: See DEFAULT_IGNORED_DIRECTORIES and DEFAULT_IGNORED_EXTENSIONS in Listen::Silencer
+                                                # default: See DEFAULT_IGNORED_FILES and DEFAULT_IGNORED_EXTENSIONS in Listen::Silencer
 
 ignore!: %r{/foo/bar}                           # Same as ignore options, but overwrite default ignored paths.
 
--- a/lib/listen/silencer.rb
+++ b/lib/listen/silencer.rb
@@ -3,8 +3,8 @@
 module Listen
   class Silencer
     # The default list of directories that get ignored.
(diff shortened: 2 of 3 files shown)


lib/listen/silencer.rb
```diff
@@ -34,6 +34,9 @@ class Silencer
       | \.swpx
       | ^4913
 
+      # Emacs backup/swap files
+      | (?:\.\#.+|\#.+\#)
```
ColinDKelley, line 38: I believe these should be anchored per our discussion on the Issue. And \A and \z are the best way to anchor in Ruby so as not to be tripped up by newlines (which are rare in filenames...but possible). So I think this does it:
```
(?:\A\.\#.|\A\#.+\#\z)
```

zw963: @ColinDKelley , oh, i add to wrong place, i thought where i insert in regexp is around with a `^...$`

for \A, \z, agree with you, but why i can't find those anchor within exists regexp? (only see some `^`, `$` anchor out of group)

ColinDKelley: I'm guessing that those who wrote the earlier code didn't know about the Ruby trap of `^` and `$` not working the same as in every other language.

I can also believe the some of the existing patterns weren't anchored at all, even when they should have been.

zw963: okay, so, double confirm, you want me fix those exists regexp anchor, and then add emacs pattern into it, right?

ColinDKelley: Yes, two separate commits should make it clear:

1. Fix the existing ^ and $ anchors to \A and \z.
2. Add your new emacs patterns with \A and \z anchors.

Or if you'd rather keep this PR simple, we can flip the order and I can take care of (2):

1. You add your new emacs patterns with ^ and $ anchors.
2. [after merging this PR] I fix all the ^ and $ anchors to \A and \z.

zw963: Got it, let me do both of them first anyway.

README.md
```diff
@@ -122,7 +122,7 @@ listener.stop    # stop both listening to changes and processing them
 
 ### Ignore / ignore!
 
-`Listen` ignores some directories and extensions by default (See DEFAULT_IGNORED_DIRECTORIES and DEFAULT_IGNORED_EXTENSIONS in Listen::Silencer).
+`Listen` ignores some directories and extensions by default (See DEFAULT_IGNORED_FILES and DEFAULT_IGNORED_EXTENSIONS in Listen::Silencer).
```
ColinDKelley, line 125:
```
`Listen` ignores some files and extensions by default (See
```

README.md
```diff
@@ -157,7 +157,7 @@ All the following options can be set through the `Listen.to` after the directory
 
 ```ruby
 ignore: [%r{/foo/bar}, /\.pid$/, /\.coffee$/]   # Ignore a list of paths
-                                                # default: See DEFAULT_IGNORED_DIRECTORIES and DEFAULT_IGNORED_EXTENSIONS in Listen::Silencer
+                                                # default: See DEFAULT_IGNORED_FILES and DEFAULT_IGNORED_EXTENSIONS in Listen::Silencer
```
ColinDKelley, line 160: Thanks for renaming this!

ColinDKelley (approved):
Just one comment update I can take care of when I merge this PR.

Thank you!

Issue #548, opened 2021-11-05
Error when renaming folder


simon1tan:
Renamed a folder and received this error:

{:modified=>[], :added=>[], :removed=>["/mnt/data/services/auto/content/1"]}
E, [2021-10-31T18:33:04.468750 #12499] ERROR -- : Exception rescued in listen-run_thread:
NoMethodError: undefined method `key?' for 1636153712.0:Float
/var/lib/gems/2.7.0/gems/listen-3.7.0/lib/listen/record.rb:58:in `block in dir_entries'
/var/lib/gems/2.7.0/gems/listen-3.7.0/lib/listen/record.rb:56:in `transform_values'
/var/lib/gems/2.7.0/gems/listen-3.7.0/lib/listen/record.rb:56:in `dir_entries'
/var/lib/gems/2.7.0/gems/listen-3.7.0/lib/listen/directory.rb:12:in `scan'
/var/lib/gems/2.7.0/gems/listen-3.7.0/lib/listen/change.rb:60:in `invalidate'
/var/lib/gems/2.7.0/gems/listen-3.7.0/lib/listen/adapter/base.rb:108:in `_queue_change'
/var/lib/gems/2.7.0/gems/listen-3.7.0/lib/listen/adapter/linux.rb:66:in `_process_event'
/var/lib/gems/2.7.0/gems/listen-3.7.0/lib/listen/adapter/base.rb:44:in `block (2 levels) in configure'
/var/lib/gems/2.7.0/gems/rb-inotify-0.10.1/lib/rb-inotify/notifier.rb:217:in `block in watch'
/var/lib/gems/2.7.0/gems/rb-inotify-0.10.1/lib/rb-inotify/watcher.rb:41:in `callback!'
/var/lib/gems/2.7.0/gems/rb-inotify-0.10.1/lib/rb-inotify/event.rb:128:in `callback!'
/var/lib/gems/2.7.0/gems/rb-inotify-0.10.1/lib/rb-inotify/notifier.rb:264:in `block in process'
/var/lib/gems/2.7.0/gems/rb-inotify-0.10.1/lib/rb-inotify/notifier.rb:263:in `each'
/var/lib/gems/2.7.0/gems/rb-inotify-0.10.1/lib/rb-inotify/notifier.rb:263:in `process'
/var/lib/gems/2.7.0/gems/rb-inotify-0.10.1/lib/rb-inotify/notifier.rb:236:in `block in run'
[...]

PR #552, opened 2021-12-30
Issue #548: fix error when renaming folder

ColinDKelley:


lib/listen/record.rb
```diff
@@ -46,16 +46,18 @@ def file_data(rel_path)
     end
 
     def dir_entries(rel_path)
-      subtree = if ['', '.'].include? rel_path.to_s
+      rel_path_s = rel_path.to_s
+      subtree = if ['', '.'].include?(rel_path_s)
```
ioquatix, line 50: I would suggest moving the assignment into the expression for readability sake.

lib/listen/record.rb
```diff
@@ -46,16 +46,18 @@ def file_data(rel_path)
...
-        @tree[rel_path.to_s]
+        @tree[rel_path_s] ||= _auto_hash
       end
 
-      subtree.transform_values do |values|
-        # only get data for file entries
-        values.key?(:mtime) ? values : {}
+      subtree.each_with_object({}) do |(key, values), result|
+        # only return data for file entries
+        if values.respond_to?(:has_key?)
```
ioquatix, line 58: What's the reason for this extra check?

lib/listen/record.rb
ColinDKelley: Which assignment? `rel_path_s =`?

lib/listen/record.rb
ColinDKelley: The root of this bug is that for folders the subtree has an array of hashes (that respond to `has_key?`). But for files, `subtree` is itself a hash. So if a folder gets renamed into a file, we were crashing here trying to treat a float (mtime) as if it were a hash.
So my hack here is to just treat the symptoms and skip over entries like that, if they happen. I'm sure there's a better answer, and it probably involves the comment higher up about deleting this entire class. But I wasn't able to take on a scope like that.

ColinDKelley: Hmm, I think I could address the above more cleanly a few lines up. Maybe check if `subtree` is an Array at all? I _think_ in the file case it will itself be a Hash.

ColinDKelley: Turns out my check for Array wouldn't work, because in fact the top-level object is a hash with the filename as key. The values in turn are hashes with symbols as keys (like `:mtime`) and scalar values.

lib/listen/record.rb
ioquatix: I would personally prefer:

```
if ...
  subtree = ... 
else
  subtree = ...
end
```

I think it's easier to read and less likely to cause bugs in the future.

lib/listen/record.rb
ioquatix: I personally think `respond_to?` is a hack in this case. Not sure if there is a better alternative without spending some time debugging the data. Maybe you can `pp` the top level out so we can see exactly what it looks like?

lib/listen/record.rb
ColinDKelley: Hmm. To me, that would be a surprising step away from DRY code. I've found the common assignment can be less bug-prone, for example if an `elsif` is added. And it often identifies a good refactor point where the entire `if`/`else` cascade can move into a separate method. I'd rather leave this for now.

lib/listen/record.rb
ColinDKelley: Agreed--it's definitely a hack!

I certainly could replace `respond_to?` with `is_a?(Hash)`. Would that seem any less hacky to you? Over the years I've gone back and forth between asking objects the `is_a?` question vs the `respond_to?`. The latter is a bit more "duck-typey" but sometimes less clear.

ioquatix: I'm guessing `transform_values` is more efficient too.

ColinDKelley: Yes, `transform_values` is probably a bit more efficient. I'd be shocked if there were a case where it mattered here, though. The real cost of this gem is the multiple threads it creates and all the event reprocessing it does.

ioquatix:
Looking good.

PR #554, opened 2022-04-08
Add Ruby 3.0/3.1 to CI

ybiquitous:
This change just adds Ruby [3.0](https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/) and [3.1](https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/) to the CI matrix.

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -18,6 +18,8 @@ jobs:
           - 2.5
           - 2.6
           - 2.7
+          - "3.0" # need quotes for "3.0"
+          - 3.1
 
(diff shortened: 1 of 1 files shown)


.github/workflows/development.yml
```diff
@@ -18,6 +18,8 @@ jobs:
           - 2.5
           - 2.6
           - 2.7
+          - "3.0" # need quotes for "3.0"
```
ybiquitous, line 21: See https://github.com/actions/runner/issues/849

PR #555, opened 2022-04-09
Disable fail-fast for CI matrix

ybiquitous:
This change aims to prevent canceling running jobs if some jobs fail.

See also:
- [Workflow syntax for GitHub Actions - GitHub Docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast)
- [How-to Github Actions: Build Matrix - Nicola Corti](https://ncorti.com/blog/howto-github-actions-build-matrix)

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -8,6 +8,7 @@ jobs:
     continue-on-error: ${{matrix.experimental}}
 
     strategy:
+      fail-fast: false
       matrix:
         os:
           - ubuntu


ColinDKelley:
@ybiquitous I understand this disables "fail fast". What was is the problem if we leave "fail fast" enabled?

ybiquitous:
@ColinDKelley Ah, sorry for the lack of why. I think it’s beneficial to disable fail-fast to make finding a problem with a specific OS or Ruby version easier.

But, since disabling fail-fast may increase CI time, it’s a trade-off.
Feel free to close this PR if you’re against it. 😃

ColinDKelley:
Ah, that does seem like a tradeoff. And the biggest reason to have Fail Fast enabled would be if you were waiting for build results as soon as possible. Which I never am here. So I'm good with this change!

ybiquitous:
@ColinDKelley Thank you for the review and merge! 😊

PR #563, opened 2022-10-24
Raise argument error when passing a file path

janko:
The Listen gem doesn't appear to support listening on changes of individual files, but currently if we pass a file path to `Listen.on`, the listener is successfully initialized, but it won't actually be listening on given files.

```rb
Listen.on(
  "app", # will listen
  "db/schema.sql" # won't listen
) { ... }
```

To improve clarity, we notify the user they need to pass a directory by raising an argument error if a path to file is passed in.

### DIFF
--- a/lib/listen/adapter/config.rb
+++ b/lib/listen/adapter/config.rb
@@ -16,6 +16,12 @@ module Listen
           Pathname.new(directory.to_s).realpath
         end
 
+        @directories.each do |pathname|
+          unless pathname.directory?
+            fail ArgumentError, "must be a directory: #{pathname}"
+          end
+        end
+
         @silencer = silencer
         @queue = queue
         @adapter_options = adapter_options
--- a/spec/lib/listen/adapter/config_spec.rb
+++ b/spec/lib/listen/adapter/config_spec.rb
@@ -17,6 +17,7 @@ RSpec.describe Listen::Adapter::Config do
   # Here's what may be  passed to initializer
   let(:path1) { fake_path('/real/path1', realpath: real_path1) }
   let(:path2) { fake_path('/real/path2', realpath: real_path2) }
+  let(:path3) { fake_path('/real/path3', realpath: real_path3) }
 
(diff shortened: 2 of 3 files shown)


lib/listen/adapter/config.rb
```diff
@@ -16,6 +16,12 @@ def initialize(directories, queue, silencer, adapter_options)
           Pathname.new(directory.to_s).realpath
         end
 
+        @directories.each do |pathname|
+          unless pathname.directory?
+            fail ArgumentError, "must be a directory: #{pathname}"
```
ColinDKelley, line 21: Good point, @janko. I confirmed in the documentation and sample usage that the given path(s) need to be directories.

Any chance there might be cases today that are passing in paths that aren't currently directories, but might be in the future? For example, I could imagine a case like this:

```
Listen.to("/tmp", "/usr/tmp", "/usr/local/tmp") do ...
```
Would that have worked before to catch whichever of those folders existed...but would now break?

In the worst case we could wait for the next major version bump to include this check. But we don't have one of those planned so that would be unfortunate.

If the above concern is real, we could possibly reverse the test and only raise in the event that the directory given is actually a plain file?

ColinDKelley: @janko Any thoughts on the questions above?

janko: @ColinDKelley Sorry, this has slipped off my radar, thanks for the input.

It seems that nonexisting directories are not currently supported by Listen, because the `realpath` call above fails with `Errno::ENOENT` when I to provide a directory that doesn't exist.

ColinDKelley: @janko Thanks for testing that. I checked the `realpath` docs and confirmed that it will fail if the path doesn't exist. So as you say, we don't have to worry about existing use cases passing in non-existent paths. 👍

ColinDKelley (approved):
Looks good. Thanks for submitting this PR and for the follow-up.

PR #566, opened 2023-01-05
Use the defined type to the default value of `directory`

y-yagi:
Currently, you will get Thor's deprecated message when starting the `Listen::CLI`.

```
Deprecation warning: Expected array default value for '--directory'; got "." (string).
This will be rejected in the future unless you explicitly pass the options `check_default_type: false` or call `allow_incompatible_default_type!` in your code
You can silence deprecations warning by setting the environment variable THOR_SILENCE_DEPRECATION.
```

This is due to the incorrect default value(`directory` is defined as an `array`, but the default value is a `string`).

This fixed to use the correct default value and correctly pass to the `directory` to `Listen.to`.


LouisaNikita:
我已收到你的邮件，谢谢！

ColinDKelley:
@y-yagi Can you clarify: when did this warning start? Was this always a bug but just not detected until a recent version of Thor?

y-yagi:
@ColinDKelley If my understanding is correct,  this warning has started since Thor v1.0.0. PR: https://github.com/rails/thor/pull/626 
This isn't a bug and still works well now. But, as the deprecated message describes, it won't work in the future version. So we need to fix it.

lib/listen/cli.rb
```diff
@@ -18,7 +18,7 @@ class CLI < Thor
 
     class_option :directory,
                  type:    :array,
-                 default: '.',
+                 default: ['.'],
                  aliases: '-d',
                  banner:  'The directory to listen to'
```
ColinDKelley, line 23:
```suggestion
                 banner:  'One or more directories to listen to'
```

ColinDKelley (commented):
Looks good. Just one suggestion on the banner. (I considered changing the option name too, to `directories`, but that would be an interface change, and the plural case is rare, so it doesn't seem worth it.)

lib/listen/cli.rb
y-yagi: Thanks for your review. I fixed.

PR #573, opened 2023-11-29
Rubocop workflow with GitHub actions

### LINKED ISSUE #516: Replace Hound with GitHub Action

AlexB52:
Fix #516 

Rubocop version `0.91` specified in the Gemfile isn't working for Ruby 3+ and the rake task fails locally.
Upgrading Rubocop solves the problem. 
I thought I'd take the opportunity to fix #516 since it's an open issue.

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -49,3 +49,18 @@ jobs:
       timeout-minutes: 6
       run: |
         ${{matrix.env}} bundle exec rspec
+
+  linting:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 3.2
+        bundler-cache: true
+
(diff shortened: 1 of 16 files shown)


ColinDKelley:
Thanks for the PR, @AlexB52.

PR #574, opened 2023-11-29
Replace Kernel warnings with Listen.logger

### LINKED ISSUE #572: Move `#warn` calls from Kernel to Logger
I'm looking to silence the Symlink error. 

```
** ERROR: directory is already being watched! **

Directory: %s

is already being watched through: %s

MORE INFO: #{README_URL}
```

People have mentioned you can monkey patch `Listen::Record::SymlinkDetector` class: https://github.com/rails/webpacker/issues/1990

I see that `logger` has a warn level that can be set, but some warn messages are made by Kernel

kernel warn references:
* https://github.com/guard/listen/blob/master/lib/listen/adapter.rb#L39
* https://github.com/guard/listen/blob/master/lib/listen/adapter/bsd.rb#L37
* https://github.com/guard/listen/blob/master/lib/listen/adapter/darwin.rb#L33
* https://github.com/guard/listen/blob/master/lib/listen/adapter/windows.rb#L23
* https://github.com/guard/listen/blob/master/lib/listen/record/symlink_detector.rb#L36

Are there specific reasons for using `Kernel#warn` instead of `Logger#warn`?
Can we move the calls from `Kernel` to `Logger`?

AlexB52:
Fix #572 

In case we can move these calls to Listen.logger here is the PR for the change.

### DIFF
--- a/lib/listen/adapter.rb
+++ b/lib/listen/adapter.rb
@@ -36,7 +36,7 @@ module Listen
 
       def _warn_polling_fallback(options)
         msg = options.fetch(:polling_fallback_message, POLLING_FALLBACK_MESSAGE)
-        Listen.logger.warn "[Listen warning]:\n  #{msg}" if msg
+        Listen.logger.warn(msg) if msg
       end
     end
   end
--- a/lib/listen/adapter.rb
+++ b/lib/listen/adapter.rb
@@ -36,7 +36,7 @@ module Listen
 
       def _warn_polling_fallback(options)
         msg = options.fetch(:polling_fallback_message, POLLING_FALLBACK_MESSAGE)
-        Kernel.warn "[Listen warning]:\n  #{msg}" if msg
+        Listen.logger.warn "[Listen warning]:\n  #{msg}" if msg
       end
     end
   end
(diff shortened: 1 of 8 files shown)


ColinDKelley:
@AlexB52 I'm concerned that log warnings will almost never be seen. That's why `Kernel.warn` or `Kernel#warn` was being called--so that these warnings would be visible. But clearly there are cases where people don't want to see one or more of the warnings.

I wonder if instead the warnings could be configurable in `listen`? Maybe there could be a proc that could be registered that would get to look at each warning and decide whether to show it or not?

AlexB52:
@ColinDKelley 
Not sure I understand the registration proc. Is your idea of a proc like this? 
What is the preferred way to make this globally available? 
Can you provide some pseudocode to illustrate the idea please?

```ruby
warn = ->(message) {
  return if ENV['LISTEN_GEM_DISABLE_WARNING'].present?

  Kernel.warn(message)
}
```
What about a Listen::Warning class that calls `Kernel.warn` unless a ENV variable `LISTEN_GEM_DISABLE_WARNING` is set. That would make it a bit more explicit maybe?

Something like:

```ruby
module Listen
  class Warning
    module_function
    def warn(message)
      return if ENV['LISTEN_GEM_DISABLE_WARNING'].present?

      Kernel.warn(message)
    end
  end
end
```

PR #577, opened 2024-02-15
Fix github workflows link in README

hotoolong:
The link was a 404 and has been corrected.


ColinDKelley:
Thank you for the fix, @hotoolong.

PR #579, opened 2024-02-22
Issue #572: make kernel warn configurable

ColinDKelley:
- Makes the usage of `Kernel#warn` configurable through `Listen.adapter_warn_behavior =`.
- Adds this documentation to the README on that feature:
### Adapter Warnings
If listen is having trouble with the underlying adapter, it will display warnings with `Kernel#warn` by default,
which in turn writes to STDERR.
Sometimes this is not desirable, for example in an environment where STDERR is ignored.
For these reasons, the behavior can be configured using `Listen.adapter_warn_behavior =`:
``` ruby
Listen.adapter_warn_behavior = :warn   # default (true means the same)
Listen.adapter_warn_behavior = :log    # send to logger.warn
Listen.adapter_warn_behavior = :silent # suppress all adapter warnings (nil or false mean the same)
```
Also there are some cases where specific warnings are not helpful.
For example, if you are using the polling adapter--and expect to--you can suppress the warning about it
by providing a callable object like a lambda or proc that determines the behavior based on the `message`:
``` ruby
Listen.adapter_warn_behavior = ->(message) do
  case message
  when /Listen will be polling for changes/
    :silent
  when /directory is already being watched/
    :log
  else
    :warn
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ Note: `:only` regexp patterns are evaluated only against relative **file** paths
 
 All the following options can be set through the `Listen.to` after the directory path(s) params.
 
-```ruby
+``` ruby
 ignore: [%r{/foo/bar}, /\.pid$/, /\.coffee$/]   # Ignore a list of paths
                                                 # default: See DEFAULT_IGNORED_FILES and DEFAULT_IGNORED_EXTENSIONS in Listen::Silencer
 
@@ -187,7 +187,7 @@ This is the primary method of debugging.
 
 ### Custom Logger
 You can call `Listen.logger =` to set a custom `listen` logger for the process. For example:
-```
+``` ruby
 Listen.logger = Rails.logger
 ```
 
@@ -197,7 +197,7 @@ If no custom logger is set, a default `listen` logger which logs to to `STDERR` 
 The default logger defaults to the `error` logging level (severity).
 You can override the logging level by setting the environment variable `LISTEN_GEM_DEBUGGING=<level>`.
 For `<level>`, all standard `::Logger` levels are supported, with any mix of upper-/lower-case:
-```
+``` ruby
 export LISTEN_GEM_DEBUGGING=debug # or 2 [deprecated]
 export LISTEN_GEM_DEBUGGING=info  # or 1 or true or yes [deprecated]
 export LISTEN_GEM_DEBUGGING=warn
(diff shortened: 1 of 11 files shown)


ColinDKelley:
@AlexB52, you asked over in the issue:

> It seems that https://github.com/rails/webpacker/issues/1990 will be able to configure listen in a rails initializer too right?

Yes, exactly. I've read through a couple of those issues and they were able to be closed without needing this... so I get a little worried that the feature here that would allow them to be silenced might simply mask a more general issue that still wasn't fixed. But I think that's an ok risk to take.

ColinDKelley:
> Having access to Listen gem directly allows configuration but what about people using Listen indirectly through guard for example. I'm not super familiar with guard, would people need to be able to silence these or would it be the gem using Listen to make a port for Listen configuration.

I'm not super-familiar with `guard` either. If you're using `Listen` in the same process, then you can simply configure it there, through the `Listen.adapter_warn_behavior =` interface.

But if it's a command-line interface, that interface would need to expose this `Listen` configuration as an option somewhere. One hacky way to do that would be as an environment variable, I suppose. `Listen` already has a few of those already.

AlexB52:
> Yes, exactly. I've read through a couple of those issues and they were able to be closed without needing this... so I get a little worried that the feature here that would allow them to be silenced might simply mask a more general issue that still wasn't fixed. But I think that's an ok risk to take.

This is how I currently silence these warnings in retest by monkey patching the `SymlinkDetector#_fail` which seems not ideal. https://github.com/AlexB52/retest/blob/main/lib/config/listen.rb

I'll definitely try to migrate over the `:log` or `:slient` adapter. 
I was concerned about ruby support. Retest currently supports Ruby 2.5, and uses Listen 3.2.  I see that [Listen supports Ruby 2.4]( https://github.com/guard/listen/blob/master/listen.gemspec#L35C3-L36C1) which is nice and should be straight forward.

ColinDKelley:
@AlexB52 I just pushed the 5 changes to call `adapter_warn` with specs [here](https://github.com/guard/listen/pull/579/commits/6197c92d872d2d3fe82aa1cee9ab0bd6a04f5f3a).

Notice that I left a `SymlinkDetector#warn` method--with a deprecation comment--since you mentioned having monkey patched that as a work-around. By stopping off there, I think we can avoid breaking any code like that.

AlexB52:
@ColinDKelley 

Sounds good. It's a good point that people could have monkey-patched the `#warn` method. 
I tested locally with your branch on a local repository that I know shows the symlink warnings

|configuration|expectation|result|
|---|---|---|
|LISTEN_GEM_ADAPTER_WARN_BEHAVIOR=warn| shows |ok|
|LISTEN_GEM_ADAPTER_WARN_BEHAVIOR=silent| muted |ok|
|LISTEN_GEM_ADAPTER_WARN_BEHAVIOR=log| muted |ok|
|Listen.adapter_warn_behavior = ->(message) { :warn }| shows |ok|
|Listen.adapter_warn_behavior = ->(message) { :silent }| muted |ok|
|Listen.adapter_warn_behavior = ->(message) { :log }| muted |ok|
|Listen.adapter_warn_behavior = :warn| shows |ok|
|Listen.adapter_warn_behavior = :silent| muted |ok|
|Listen.adapter_warn_behavior = :log| muted |ok|

Note: I did not try to test whether the warnings were pushed to a log file but only that logs were muted.

ColinDKelley:
That looks great, @AlexB52. Thanks for the thorough testing.

I'm good with merging it now. Sound good?

AlexB52:
@ColinDKelley yup sounds good

PR #584, opened 2024-08-05
Remove conditions from gem declarations in Gemfile

larskanis:
Both gems 'wdm' and 'rb-kqueue' can be installed on any platform. They fail when require 'wdm' or 'rb-kqueue' is called. So the conditions in the Gemfile can be removed and should not be recommended.

Having conditions in the Gemfile is generally bad practice, since this leads to changed Gemfile.lock, whenever the bundler is executed on a different platform. Also `bundle package --all-platforms` doesn't work properly, when gems are declared conditionally.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,14 +11,8 @@ else
   gemspec development_group: :gem_build_tools
 end
 
-require 'rbconfig'
-
-case RbConfig::CONFIG['target_os']
-when /mswin|mingw|cygwin/i
-  gem 'wdm', '>= 0.1.0'
-when /bsd|dragonfly/i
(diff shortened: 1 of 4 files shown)


LouisaNikita:
我已收到你的邮件，谢谢！

larskanis:
In principle the gems could also be added to the `listen.gemspec`, like 'rb-fsevent' and
'rb-inotify'. But this would install them for all users of 'listen', although only a small subset needs them.

larskanis:
Any chance to get this merged?

rymai:
Thanks, looks good to me!

PR #586, opened 2025-01-17
Remove outdated note about performance

quadule:
Silenced directories aren't recorded since https://github.com/guard/listen/pull/544, so ignoring large directories like `.git` and `node_modules` can improve performance.


LouisaNikita:
我已收到你的邮件，谢谢！

PR #587, opened 2025-01-19
Add `logger` gem as a runtime dependency

y-yagi:
`logger` will be a bundled gem from Ruby 3.5.
https://bugs.ruby-lang.org/issues/20309

So if we use `logger` as the standard library, Bundler shows the following warning.

```
/home/y-yagi/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/listen-3.9.0/lib/listen.rb:3: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
```

This PR adds the `logger` to runtime dependency to fix the warning.


y-yagi:
The linting is failed. I will fix it with #588.

rymai:
Thanks!

y-yagi:
@rymai 
Sorry to ping you, but could you release a new gem including this fix? Ruby 4.0 has released and this fix is needed for Ruby 4.0.

LouisaNikita:
我已收到你的邮件，谢谢！

PR #590, opened 2025-08-06
CI against Ruby 3.4

y-yagi:
This PR just adds the Ruby 3.4 to the CI matrix to check `listen` works with it.


rymai:
@y-yagi Thanks, looks good to me! 🚀

PR #596, opened 2026-01-05
CI against Ruby 4.0

y-yagi:
This PR just adds the Ruby 4.0 to the CI matrix to check listen works with it.


rymai:
Thanks, looks good to me! ❤️

PR #598, opened 2026-08-19
[Darwin] Use FSEvents when `target_os` has no version number

klizas:
On Ruby configured with `--with-os-version-style=none`, `RbConfig::CONFIG['target_os']` is a bare `darwin` rather than e.g. `darwin24`. `OS_REGEXP` required `(1|2)\d+`, so `usable?` returned false, and since no other adapter is usable on macOS, Listen fell back to polling.

```
[Listen warning]:
  Listen will be polling for changes.Learn more at https://github.com/guard/listen#listen-adapters.
```

Distributors use that configure option so `RUBY_PLATFORM` stays stable across macOS releases, where it is otherwise derived from the build machine's kernel. Either way `target_os` is a cosmetic build string, so it says nothing about whether FSEvents is available.

This PR makes the version optional and treats an absent version as usable. The `rb-fsevent` check still runs when a version is present and older than darwin13, so OS X 10.6-10.8 is unaffected. `\A` is needed because an optional capture group would otherwise let the pattern match `darwin` anywhere in the string. As a side effect the pattern now also matches `darwin30` and later, which `(1|2)\d+` did not.


LouisaNikita:
我已收到你的邮件，谢谢！

LouisaNikita:
我已收到你的邮件，谢谢！

